PR #128, opened 2020-11-20
Fix #126. Running jupyterhub under Sirepo.

### LINKED ISSUE #126: Support running jupyterhub under Sirepo

e-carlin:
I believe this covers everything for running jupyterhub at sirepo.com/jupyter.

For dev getting everything running on a remote server using ssh LocalForward required some hacks. There were issues with re-directs dropping the port number and jupyterhub blocking requests due to them being cross origin. In the end I setup a reverse proxy on my mac and edited /etc/hosts to make it all work. My nginx foo isn't the best so maybe there is a simpler way to get it working that doesn't require a reverse proxy.

~/.ss/config on mac:

Host <redacted>
  HostName <redacted>
  Port  <redacted>
  User <redcated>
  LocalForward 8080 sirepo.v4.radia.run:443

nginx on mac:

brew install nginx
cd /usr/local/etc/nginx
sudo openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout cert.key -out crt.pem # for fqdn use sirepo.v4.radia.run
cat <<EOF > /usr/local/etc/nginx/nginx.conf
worker_processes  1;
error_log /usr/local/var/log/nginx/error.log debug;

events {
    worker_connections  1024;
}

http {
    include       mime.types;
    default_type  application/octet-stream;

    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for"';

    sendfile        on;
    keepalive_timeout  65;

    proxy_buffer_size   128k;
    proxy_buffers   4 256k;
    proxy_busy_buffers_size   256k;
    server {
      listen 443 ssl;
      server_name sirepo.v4.radia.run;
      ssl_certificate      cert.pem;
      ssl_certificate_key  cert.key;
      proxy_http_version 1.1;
      proxy_next_upstream error timeout invalid_header http_500 http_502 http_503 http_504;
      proxy_read_timeout 600s;
      proxy_send_timeout 600s;
      proxy_connect_timeout 600s;
      proxy_redirect off;
      proxy_set_header Host $host:$server_port;
      proxy_set_header Proxy "";
      proxy_set_header Via $remote_addr;
      proxy_set_header X-Real-Ip $remote_addr;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header X-Forwarded-Proto $scheme;
      proxy_cache_bypass $http_upgrade;
      proxy_set_header Upgrade $http_upgrade;
      proxy_set_header Connection "upgrade";
      location / {
          proxy_pass         https://localhost:8080;
      }
    }
    include servers/*;
}
EOF
brew services start nginx

/etc/hosts on mac:
127.0.0.1 localhost localhost.localdomain sirepo.v4.radia.run

Then visit https://sirepo.v4.radia.run/jupyter in your browser.

### DIFF
--- a/rsconf/component/sirepo.py
+++ b/rsconf/component/sirepo.py
@@ -187,6 +187,9 @@ class T(component.T):
             self.install_directory(p.join(c))
         self.install_access(mode='400')
         for c in z.feature_config.proprietary_sim_types:
+            # TODO(e-carlin): Need a better way to configure this
+            if c == 'jupyterhublogin':
+                continue
             self.install_abspath(
                 self.rpm_file(jc, f'{_RPM_PREFIX}{c}'),
                 p.join(c, c + '.rpm'),
--- a/rsconf/component/sirepo.py
+++ b/rsconf/component/sirepo.py
@@ -36,6 +36,9 @@ class T(component.T):
         from rsconf.component import docker_registry
         from rsconf import db
 
+        self.__docker_unit_enable_after = []
+        self.__uwsgi_docker_vols = []
+
         self.buildt.require_component('docker', 'nginx', 'db_bkp')
         jc, z = self.j2_ctx_init()
         self.__run_d = systemd.docker_unit_prepare(self, jc)
@@ -56,6 +59,7 @@ class T(component.T):
                 feature_config=dict(
                     api_modules=[],
                     job=True,
+                    other_sim_types=tuple(),
                     proprietary_sim_types=tuple(),
                 ),
                 job=dict(
@@ -78,6 +82,7 @@ class T(component.T):
             ),
         ))
         self._comsol(z)
+        self._jupyterhublogin(z)
         self.j2_ctx_pksetdefault(dict(
             sirepo=dict(
                 client_max_body_size=pkconfig.parse_bytes(z.job.max_message_bytes),
@@ -107,10 +112,7 @@ class T(component.T):
             z.pkcli.job_supervisor.ip,
             z.pkcli.job_supervisor.port,
         )
-        self.buildt.require_component('sirepo_job_supervisor')
-        s = self.buildt.get_component('sirepo_job_supervisor')
-        s.sirepo_config(self)
-        self.__docker_unit_enable_after = [s.name]
+        self._set_sirepo_config('sirepo_job_supervisor')
 #TODO(robnagler) remove when deployed to alpha
         z.job.supervisor_uri = z.job_api.supervisor_uri
 
@@ -122,12 +124,18 @@ class T(component.T):
         jc = self.j2_ctx
         z = jc[self.name]
         self._install_dirs_and_files()
+        k = PKDict()
+        if self._jupyterhublogin_enabled():
+            k.jupyterhub_enabled = True
+            k.jupyterhub_proxy_port = jc.sirepo_jupyterhub.port
+            k.jupyterhub_uri_root = jc.sirepo_jupyterhub.uri_root
         nginx.install_vhost(
             self,
             vhost=z.vhost,
             backend_host=jc.rsconf_db.host,
             backend_port=z.pkcli.service_port,
             j2_ctx=jc,
+            nginx_kwargs=k,
         )
         systemd.docker_unit_enable(
             self,
@@ -137,6 +145,7 @@ class T(component.T):
             cmd='sirepo service uwsgi',
             after=self.__docker_unit_enable_after,
             #TODO(robnagler) wanted by nginx
+            volumes=self.__uwsgi_docker_vols,
         )
         db_bkp.install_script_and_subdir(
             self,
@@ -189,3 +198,31 @@ class T(component.T):
                 self.rpm_file(jc, f'{_RPM_PREFIX}{c}'),
                 p.join(c, c + '.rpm'),
             )
+
+    def _jupyterhublogin(self, z):
+        if not self._jupyterhublogin_enabled():
+            return
+        self.buildt.require_component('jupyterhub')
+        j = self.buildt.get_component('jupyterhub')
+        d = j.jc.jupyterhub.user_d
+        self.j2_ctx_pksetdefault(dict(
+            sirepo=dict(
+                sim_api=dict(
+                    jupyterhublogin=dict(
+                        src_db_root=d,
+                        dst_db_root=d,
+                    ),
+                ),
+            ),
+        ))
+        self.__uwsgi_docker_vols.append(d)
+        self._set_sirepo_config('sirepo_jupyterhub')
+
+    def _jupyterhublogin_enabled(self):
+        return 'jupyterhublogin' in self.j2_ctx.sirepo.feature_config.other_sim_types
+
+    def _set_sirepo_config(self, component):
+        self.buildt.require_component(component)
+        c = self.buildt.get_component(component)
+        c.sirepo_config(self)
+        self.__docker_unit_enable_after.append(c.name)
--- a/rsconf/component/sirepo.py
+++ b/rsconf/component/sirepo.py
@@ -202,20 +202,8 @@ class T(component.T):
     def _jupyterhublogin(self, z):
         if not self._jupyterhublogin_enabled():
             return
-        self.buildt.require_component('jupyterhub')
-        j = self.buildt.get_component('jupyterhub')
-        d = j.jc.jupyterhub.user_d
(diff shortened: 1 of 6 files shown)


e-carlin:
@robnagler I've addressed all comments. Ready for another review.

robnagler:
Is there a missing commit in container-jupyterhub?

```txt
sirepo_jupyterhub[7968]: ModuleNotFoundError: No module named 'sirepo'
```

I figure it must be a dependency now  in the build.sh, but maybe that wasn't committed?

rsconf/component/sirepo_jupyterhub.py
```diff
@@ -86,6 +86,8 @@ def _env_ok(elem):
         self.install_access(mode='711', owner=z.run_u)
         self.install_directory(self.__run_d)
         self.install_access(mode='700', owner=z.jupyter_run_u)
+        if z.user_d.dirname == '/srv/jupyterhub':
+            self.install_directory(z.user_d.dirname)
         self.install_directory(z.user_d)
```
e-carlin, line 91: Should this line be deleted?

e-carlin: Disregard. I see what's going on.

e-carlin:
I was using radiasoft/sirepo as the base image in my testing but adding Sirepo as a dep makes considerably more sense. Can you give me access to container-jupyterhub?

e-carlin:
Just adding Sirepo as a dep doesn't work because some imports are missing (ex h5py). How should we resolve? I think using sirepo as the base image is overkill because the image includes all of the codes which jupyterhub doesn't need. OTOH using it as the base image will always means sirepo works...

robnagler:
Does sirepo need h5py if there are no codes imported?

robnagler:
I gave you access

e-carlin:
Thanks.

No [h5py isn't imported in that case](https://github.com/radiasoft/sirepo/blob/3b1657c974f1f64768b4861473f1fc0b02d56113/sirepo/server.py#L37).

robnagler:
I think we should import only those sim_types that are necessary, possibly just jupyterlogin.  The configuration needs to be custom, which we can override inside the build.sh, since it will always be true, not just in rsconf. Some config can't be overriden (secret, for example).

e-carlin:
Ok. You lost me on what build.sh has to do with this? Maybe I wasn't clear above but the image builds fine. It is just when running jupyterhub and it imports sirepo then sirepo imports h5py that the problem arises. So, I think we can set feature config sim_types to be jupyterhublogin in sirepo_jupyterhub's env and all should work fine. Agree?

robnagler:
I see that we don't have a radia-run.sh in container-jupyterhub like we do with sirepo. This means we don't have a hook to set the environment in which jupyter starts up. This has to be in sirepo_juptyerhub, then. It's probably fine, but you'll also need to modify test.sh to pass SIREPO_FEATURE_CONFIG_SIM_TYPES.

e-carlin:
@robnagler I update the sirepo_jupyterhub env in rsconf to set SIREPO_FEATURE_CONFIG_SIM_TYPES. 

I realized container-jupyterhub/container-config/test.sh doesn't need to have the env. It uses a dummy authenticator so Sirepo is never imported.

rsconf/package_data/dev/db/000.yml.jinja
```diff
@@ -64,9 +64,6 @@ default:
           on_calendar: "1 0"
           bash_script: |
             echo doing that thing
-
```
robnagler, line 67: needs one blank line after a section that begins with `|`

rsconf/package_data/dev/db/000.yml.jinja
```diff
@@ -240,6 +244,13 @@ channel:
         mail_server: {{ host }}
         mail_username: comsol@{{ host }}
         mail_password: somepass
+      feature_config:
+       proprietary_sim_types:
```
robnagler, line 248: indentation incorrect

rsconf/package_data/sirepo/nginx.conf.jinja
```diff
@@ -10,6 +10,21 @@ server {
     root {{ nginx.default_root }};
     ssl_certificate {{ nginx.tls_crt }};
     ssl_certificate_key {{ nginx.tls_key }};
+{% if 'jupyterhub_enabled' in sirepo %}
```
robnagler, line 13: `{% if sirepo.jupyterhub_enabled %}`

rsconf/package_data/sirepo/nginx.conf.jinja
```diff
@@ -10,6 +10,21 @@ server {
...
+    proxy_http_version 1.1;
```
robnagler, line 14: [these are in ngnix.conf](https://github.com/radiasoft/rsconf/blob/84db636618318686e142cb0ba13448896919d50f/rsconf/package_data/nginx/global.conf.jinja#L85) so you don't need them here.

rsconf/package_data/sirepo/nginx.conf.jinja
```diff
@@ -18,4 +33,12 @@ server {
         uwsgi_send_timeout 300s;
         uwsgi_pass {{ nginx.backend_host }}:{{ nginx.backend_port }};
     }
+{% if 'jupyterhub_enabled' in sirepo %}
```
robnagler, line 36: `{% if sirepo.jupyterhub_enabled %}`

rsconf/component/sirepo_jupyterhub.py
```diff
@@ -0,0 +1,205 @@
...
+            gen_secret=lambda: db.random_string(length=64, is_hex=True),
+            visibility='channel',
+        )[0]
+        z.admin_users_str = _list_to_str(z.admin_users)
+        self._rsdockerspawner(jc, z)
+
+        self.install_access(mode='711', owner=z.run_u)
+        self.install_directory(self.__run_d)
+        self.install_access(mode='700', owner=z.jupyter_run_u)
+        if z.user_d.dirname == '/srv/jupyterhub':
```
robnagler, line 89: This is annoying. I don't have a solution for it, and we'll need it for non-sirepo jupyter servers until we solve the problem of choosing pools.

robnagler (changes requested):

rsconf/package_data/sirepo/nginx.conf.jinja
e-carlin: 361965b

rsconf/package_data/dev/db/000.yml.jinja
e-carlin: 361965b

rsconf/package_data/dev/db/000.yml.jinja
e-carlin: 361965b

rsconf/package_data/sirepo/nginx.conf.jinja
e-carlin: https://github.com/radiasoft/rsconf/pull/128/commits/0090bba595e55686c9e97eb30d38cf762f5612d2

rsconf/package_data/sirepo/nginx.conf.jinja
e-carlin: https://github.com/radiasoft/rsconf/pull/128/commits/0090bba595e55686c9e97eb30d38cf762f5612d2

rsconf/component/sirepo.py
```diff
@@ -187,6 +187,9 @@ def _install_dirs_and_files(self):
             self.install_directory(p.join(c))
         self.install_access(mode='400')
         for c in z.feature_config.proprietary_sim_types:
+            # TODO(e-carlin): Need a better way to configure this
+            if c == 'jupyterhublogin':
```
e-carlin, line 191: @robnagler how should we ask the question of whether or not a proprietary code has an rpm?

e-carlin:
@robnagler all comments addressed. One [question for you](https://github.com/radiasoft/rsconf/pull/128#discussion_r537196719).

rsconf/component/sirepo.py
robnagler: I thought I saw you called proprietary_rpm or something like that. Any interface is fine as it is quite internal.

PR #138, opened 2021-02-04
radiasoft/sirepoflash#16: Properietary codes (FLASH) are now tarballs

e-carlin:


robnagler (approved):
Excellent!

robnagler:
We have to wait to merge, I think, until we are ready to deploy sirepo to alpha.

One problem is that we will need to managing staging. We have to have a clean pipeline (no hot fixes) from alpha, beta, and prod, or we have to incorporate a conditional configuration. I'm voting for the former. @moellep what say you?

PR #141, opened 2021-03-01
move jupyterhublogin to default_proprietary_sim_types

e-carlin:
Companion pr for changes in https://github.com/radiasoft/sirepo/pull/3368

### DIFF
--- a/rsconf/component/sirepo.py
+++ b/rsconf/component/sirepo.py
@@ -186,10 +186,11 @@ class T(component.T):
             self.install_directory(p.join(c))
         self.install_access(mode='400')
         for c in z.feature_config.proprietary_sim_types:
-            self.install_abspath(
-                self.proprietary_file(jc, c),
-                p.join(c, c + '.tar.gz'),
-            )
+            if c == 'flash':
+                self.install_abspath(
+                    self.proprietary_file(jc, c),
+                    p.join(c, c + '.tar.gz'),
+                )
 
     def _jupyterhublogin(self, z):
         z.jupyterhub_enabled =  'jupyterhublogin' in self.j2_ctx.sirepo.feature_config.default_proprietary_sim_types
--- a/rsconf/component/sirepo.py
+++ b/rsconf/component/sirepo.py
@@ -57,6 +57,7 @@ class T(component.T):
                 feature_config=dict(
                     api_modules=[],
                     job=True,
+                    default_proprietary_sim_types=tuple(),
                     proprietary_sim_types=tuple(),
                 ),
                 job=dict(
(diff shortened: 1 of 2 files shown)


rsconf/component/sirepo.py
```diff
@@ -185,14 +186,13 @@ def _install_dirs_and_files(self):
...
-                    p.join(c, c + '.tar.gz'),
-                )
+            self.install_abspath(
+                self.proprietary_file(jc, c),
+                p.join(c, c + '.tar.gz'),
+            )
 
     def _jupyterhublogin(self, z):
-        z.jupyterhub_enabled =  'jupyterhublogin' in self.j2_ctx.sirepo.feature_config.proprietary_sim_types
+        z.jupyterhub_enabled =  'jupyterhublogin' in self.j2_ctx.sirepo.feature_config.default_proprietary_sim_types
```
robnagler, line 195: i think this needs to be proprietary_sim_types here, too, because it is a proprietary_sim_type and it is a default_proprietary_sim_type. I think you could add() the two sets so you ensure that proprietary_sim_types is set properly in feature_config as well as default_proprietary_sim_types.

e-carlin: Can you elaborate? Why is it both types? In [sirepo.feature_config we assert](https://github.com/radiasoft/sirepo/blob/master/sirepo/feature_config.py#L122) that it is in one or the other. It is nice to have them separate in rsconf because then we only do the proprietary file stuff for codes in proprieatary_sim_types.

robnagler: alrighty that makes sense so ignore my comment.

rsconf/component/sirepo.py
```diff
@@ -185,14 +186,13 @@ def _install_dirs_and_files(self):
             self.install_directory(p.join(c))
         self.install_access(mode='400')
         for c in z.feature_config.proprietary_sim_types:
-            if c != 'jupyterhublogin':
```
robnagler, line 188: I still think this isn't quite right. There should be some feature test or config that a proprietary_code has proprietary_file. I think this should be `if c == 'flash'` then there's a file a to copy. It doesn't have to be a config for now, because we only have such code.

https://github.com/radiasoft/sirepo/issues/3376 discusses the general problem.

robnagler (changes requested):

rsconf/component/sirepo.py
e-carlin: https://github.com/radiasoft/rsconf/pull/141/commits/5469508058d2d9b78eae75e580d6b2913c09bcc6

e-carlin:
@robnagler all comments addressed. Ready for another review.

robnagler (approved):

PR #183, opened 2022-04-13
fix #182: add sirepo_uri to sirepo_jupyterhub

### LINKED ISSUE #182: sirepo_uri for sirepo_jupyterhub
cc: @rorour (can't figure out how to get your name to appear in the list of assignees...)

@robnagler can you provide some guidance on how to get the sirepo uri so it can be provided to sirepo_jupyterhub for https://github.com/radiasoft/sirepo/commit/78abea726afdd196a064a36d87dacf5c52bc0b0a ?

I think it might be from the sirepo.vhost but not positive. Also, can we just assume https? [Looks like it](https://github.com/radiasoft/rsconf/blob/ae704a1eb3c4141bfa7620d0b048bb3d5eafd36b/rsconf/component/sirepo_job_supervisor.py#L99) but I want to confirm

e-carlin:


rsconf/component/sirepo.py
```diff
@@ -57,7 +57,7 @@ def internal_build_compile(self):
                 feature_config=PKDict(
                     api_modules=[],
                     job=True,
-                    default_proprietary_sim_types=tuple(),
+                    moderated_sim_types=tuple(),
```
robnagler, line 60: I think this needs to be backward compatible so we can use rsconf during the transition from alpha to prod. If default_proprietary_sim_types is set, use old behavior (Authenticator) otherwise use use new behavior (SirepoAuthenticator).

rsconf/component/sirepo_jupyterhub.py
```diff
@@ -18,6 +18,10 @@ class T(rsconf.component.jupyterhub.T):
     def internal_build_compile(self):
         # TODO(e-carlin): should this come from somewhere or ok to hardcode?
         self.__jupyterhub_run_d = pkio.py_path('/srv/jupyterhub')
+
+        # Must set directly on hdb so it is available in
+        # rsconf.component.jupyterhub.T.internal_build_compile
+        self.hdb[self.name].sirepo_uri = f'https://{self.hdb.sirepo.vhost}'
```
robnagler, line 24: This should be set in sirepo_config, because it is the point when you know sirepo is passing its config explicitly to sirepo_jupyterhub.

robnagler (approved):
Backwards compatibility is the big item.

rsconf/component/sirepo_jupyterhub.py
e-carlin: c9cbcdc

rsconf/component/sirepo.py
e-carlin: https://github.com/radiasoft/rsconf/pull/183/commits/72f514ded748f8ad85c8781c3da339dda93c2f5a

e-carlin:
@robnagler ready for another review.

rsconf/package_data/sirepo_jupyterhub/conf.py.jinja
```diff
@@ -26,7 +26,8 @@ c.DockerSpawner.image = '{{ sirepo_jupyterhub.jupyter_docker_image }}'
 c.DockerSpawner.image_whitelist = []
 c.DockerSpawner.network_name = 'host'
 c.DockerSpawner.use_internal_ip = True
-c.JupyterHub.authenticator_class = sirepo.jupyterhub.Authenticator
+c.JupyterHub.authenticator_class = getattr(sirepo.jupyterhub, '{{ sirepo_jupyterhub.authenticator_class }}')
+c.SirepoAuthenticator.sirepo_uri = '{{ sirepo_jupyterhub.sirepo_uri }}'
```
robnagler, line 30: I don't think this will work, because it may not be defined so better to use jinja:

```txt
{% if 'moderated_sim_types' in sirepo_jupyterhub %}
c.JupyterHub.authenticator_class = sirepo.jupyterhub.SirepoAuthenticator
c.SirepoAuthenticator.sirepo_user = '{{ sirepo_jupyterhub.sirepo_uri }}'
{% else %}
{%#TODO remove #%}
c.JupyterHub.authenticator_class = sirepo.jupyterhub.Authenticator
{% endif %}
```

rsconf/component/sirepo.py
```diff
@@ -194,7 +197,8 @@ def _install_dirs_and_files(self):
             )
 
     def _jupyterhublogin(self, z):
-        z.jupyterhub_enabled =  'jupyterhublogin' in self.j2_ctx.sirepo.feature_config.default_proprietary_sim_types
+        z.jupyterhub_enabled =  'jupyterhublogin' in set(self.j2_ctx.sirepo.feature_config.moderated_sim_types).union(
```
robnagler, line 200: I don't think this will work. default_proprietary_sim_types will go away eventually so the code must be robust to that. use `get`.

rsconf/component/sirepo_jupyterhub.py
```diff
@@ -23,6 +23,15 @@ def internal_build_compile(self):
     def sirepo_config(self, sirepo):
         self.j2_ctx_pksetdefault(sirepo.j2_ctx)
         sirepo.j2_ctx.sirepo_jupyterhub.hub_ip = self.j2_ctx.sirepo_jupyterhub.hub_ip
+        self.j2_ctx.sirepo_jupyterhub.sirepo_uri = f'https://{sirepo.j2_ctx.sirepo.vhost}'
+        c = 'SirepoAuthenticator'
+        p = sirepo.j2_ctx.sirepo.feature_config.get('default_proprietary_sim_types', [])
```
robnagler, line 28: Might be good to combine this and the other union, and just have a boolean for this.

robnagler (changes requested):

rsconf/package_data/sirepo_jupyterhub/conf.py.jinja
e-carlin: 87e5cf5

rsconf/component/sirepo_jupyterhub.py
e-carlin: 87e5cf5

rsconf/component/sirepo.py
e-carlin: 87e5cf5

robnagler (approved):

PR #208, opened 2022-07-04
WIP radiasoft/download#233: env comes from env_keep

e-carlin:
Previously, env was hardcoded in the kernel.json (set at build time).
This doesn't work because some values (ex LD_LIBRARY_PATH) are only
known at runtime.

Unfortunately, jupyterhub doesn't allow for augmenting of the default env_keep
values. Some of them are required for the notebook to work so, they are copied
here and augmented.

### DIFF
--- a/rsconf/package_data/jupyterhub/conf.py.jinja
+++ b/rsconf/package_data/jupyterhub/conf.py.jinja
@@ -42,6 +42,19 @@ c.RSDockerSpawner.cfg = '''{{ jupyterhub.rsdockerspawner_cfg }}'''
 c.JupyterHub.spawner_class = rsdockerspawner.RSDockerSpawner
 if hasattr(rsdockerspawner.RSDockerSpawner, 'sirepo_template_dir'):
     c.JupyterHub.template_paths = [rsdockerspawner.RSDockerSpawner.sirepo_template_dir()]
+# Relevant default values from jupyter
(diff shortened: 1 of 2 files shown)


robnagler:
As noted in https://github.com/radiasoft/download/pull/257#discussion_r914085974 I think this is unnecessary.

e-carlin:
Hmm I think this is still needed. We source bashrc but that env isn't carried to the Jupyter notebook unless we explicitly tell it the values that should be carried over. That is what we are doing here. 

W/o this change if you start a notebook and run 
```
import os 
os.environ['LD_LIBRARY_PATH']
```

You'll get a key error which I think is the wrong behavior. 

Thoughts?

e-carlin:
After reading the jupyterhub documentation more closely. I think my confusion is that I'm using a singleuserserver in dev and this conf is only needed for it. Sorry for my confusion. 

I'm going to verify this isn't needed with how we configure in higher environments and close.

robnagler:
Perhaps we should fix singleuserserver and/or upgrade to a newer jupyterhub (if we can) that changes the way the server starts.

e-carlin:
I'll look for a better option than singleuserserver. We are on jupyterhub 1.4.2 and the most recent is 2.3.1 so probably worth upgrading at some point.

PR #217, opened 2022-07-26
Fix #210: create sirepo_sandbox

### LINKED ISSUE #210: sirepo_sandbox component
A sirepo_sandbox is almost like mpi_worker: a docker container that runs sshd for a single user. The purpose is to allow a user to develop sirepo, using ssh to copy files.

Unlike mpi_worker, there need to be multiple instances on the node. This is similar to how bop starts multiple apps. However, these users need different ports to ssh into. 

The network should be isolated so only ssh will work, and everything else is tunneled into and out of the container via ssh. This would allow a user to  connect to the development server on the node from a local browser. We will need to use docker proxy for this.

Files have to persist so there will be read-write volumes mounted, specifically ~/.ssh, ~/src/radiasoft, and ~/jupyter. The latter is used to persist ~/jupyter/pre_jupyter_bashrc and ~/jupyter/bashrc. That's a convenient hook for this problem. No need to invent something new for this problem.

Already encrypted private keys will need to be distributed to end users outside the rsconf configured nodes. This is different than mpi_worker which uses non-encrypted keys and the distribution happens to external devices (laptops). The main point is the key is encrypted by sirepo_sandbox component with a passphrase that is configured and transferred to users along with the keys. The protocol can be manual as it is all happening inside the organization.

e-carlin:
@robnagler this creates the sirepo_sandbox component. I have it working (can ssh in using identity file). But, there are 2 TODO's below that I could use your thoughts on (1. how to get /srv/ inside of the component, 2. how to mount ~/src as a volume).  

Example db/000.yml config
```yaml
sirepo_sandbox:
  docker_image: radiasoft/beamsim-jupyter
  guest_d: /home/vagrant/
  # TODO(e-carlin): this should come from rsconf (see comment in sirepo_sandbox.py)
  host_root_d: /srv/
  users:
    e-carlin: 9999
    nagler: 8888
  volumes:
    - jupyter
    # TODO(e-carlin): doesn't work (problem with home-env not existing)
    # - src
```

Example db/secret/000.yml config
```yaml
sirepo_sandbox:
  user_passwords:
    nagler: pass-nagler
      e-carlin: pass-e-carlin

### CODE REVIEW

File: rsconf/component/sirepo_sandbox.py [hunk 1]
```diff
@@ -0,0 +1,124 @@
...
+            if f.exists():
+                continue
+            subprocess.check_call(
+                [
+                    "ssh-keygen",
+                    "-q",
+                    "-t",
+                    "ed25519",
+                    "-N",
+                    z.user_passwords[self.name],
```
e-carlin, line 104: You said:

> Already encrypted private keys will need to be distributed to end users outside the rsconf configured nodes. This is different than mpi_worker which uses non-encrypted keys and the distribution happens to external devices (laptops). The main point is the key is encrypted by sirepo_sandbox component with a passphrase that is configured and transferred to users along with the keys. The protocol can be manual as it is all happening inside the organization.

I wasn't 100% sure what you meant but is this it? We can manually create passwords and populate them in db/secret/000.yml, run `rsconf build`, then manually distribute the password/key to users?

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
+# -*- coding: utf-8 -*-
```
robnagler, line 1: I think this should call this devbox. sirepo_sandbox is tool long. The instances would be devbox_ecarlin, devops_nagler, etc.

rsconf/package_data/sirepo_sandbox/sshd_config.jinja
```diff
@@ -0,0 +1,10 @@
+AuthorizedKeysFile {{ sirepo_sandbox.guest.identity_pub_f }}
+HostKey {{ sirepo_sandbox.guest.host_key_f }}
+ListenAddress {{ sirepo_sandbox.ip }}:{{ sirepo_sandbox.ssh_port }}
+PasswordAuthentication no
+PermitTunnel no
```
robnagler, line 5: Not needed.

rsconf/package_data/sirepo_sandbox/sshd_config.jinja
```diff
@@ -0,0 +1,10 @@
...
+PermitUserEnvironment no
```
robnagler, line 6: mpi_worker again, we are trying to control things so mpi works. for "normal" logins, this is fine.

rsconf/package_data/sirepo_sandbox/sshd_config.jinja
```diff
@@ -0,0 +1,10 @@
...
+PrintLastLog no
+PrintMotd no
```
robnagler, line 8: We do want this, because it contains a disclaimer. mpi_worker is different. We want users to think the are logging into a "real" machine.

rsconf/package_data/sirepo_sandbox/sshd_config.jinja
```diff
@@ -0,0 +1,10 @@
...
+Protocol 2
+UseDNS no
```
robnagler, line 10: I don't think we need this. People will be logging in from real computers on the internet.

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+            self.install_directory(d)
+        self.install_access(mode="400")
+        self.install_resource(z.host.sshd_config, jc)
+        for k, v in z.secrets.items():
+            self.install_abspath(v, z.host[k])
+
+    def _network(self, jc, z):
+        n = self.buildt.get_component("network")
+        z.ip, _ = n.ip_and_net_for_host(jc.rsconf_db.host)
+        z.ssh_port = jc.sirepo_sandbox.users[self.name]
```
robnagler, line 79: I think self.name will be devbox_<name>. I think we have to cascade a different key (user_name) for this.

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+        z = jc.sirepo_sandbox
+        z.setdefault("volumes", {})
+        # TODO(e-carlin): Better way then building the path manually
+        # /srv/<component-name> is output by systemd.docker_unit_prepare
+        # below but to call that we need z.host.ssh_d. z.host needs z.host_d
+        # so we are in a dependency cycle
+        z.host_d = z.host_root_d.join(self.name, "db")
+        z.secrets = self._gen_keys(jc, z)
+        for x in "guest", "host":
+            z[x] = self._gen_paths(z, z.get(x + "_d"))
```
robnagler, line 37: fail fast: `z[x + "_d']`

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+                    str(f),
+                ],
+                stderr=subprocess.STDOUT,
+                shell=False,
+            )
+        return res
+
+    def _gen_paths(self, z, db_d):
+        res = PKDict()
+        s = db_d.join(".ssh")
```
robnagler, line 118: I don't think we want this as a dot file. This could be called sshd and then mounted in the guest that way.

rsconf/component/sirepo_sandbox.py
robnagler: I think we can generate passwords and it will be simpler than configuring them. They do not need to be distributed, just dumped to json. Just like rsconf does, for example. These passwords will be changed by the user once the private key is on their machine.

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+        z.ip, _ = n.ip_and_net_for_host(jc.rsconf_db.host)
+        z.ssh_port = jc.sirepo_sandbox.users[self.name]
+
+    def _gen_keys(self, jc, z):
+        from rsconf import db
+
+        res = PKDict()
+        b = db.secret_path(
+            jc,
+            "sirepo_sandbox/" + self.name,
```
robnagler, line 87: I don't think this needs to be on the channel, it can be just on the host. There's no channel relationship that needs to be maintained. Fix secret_path to take a parameter "directory=True" in this case that would do a mkdir_parent instead of a mkdir_parent_only. This is probably used in other places.

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+    def _network(self, jc, z):
+        n = self.buildt.get_component("network")
+        z.ip, _ = n.ip_and_net_for_host(jc.rsconf_db.host)
+        z.ssh_port = jc.sirepo_sandbox.users[self.name]
+
+    def _gen_keys(self, jc, z):
+        from rsconf import db
+
+        res = PKDict()
+        b = db.secret_path(
```
robnagler, line 85: share all this with mpi_worker in `component.__init__`

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+class T(component.T):
+    def install_resource(self, *args, **kwargs):
+        super().install_resource(*args, component_name="sirepo_sandbox", **kwargs)
+
+    def internal_build_compile(self):
+        from rsconf import systemd
+
+        if self.name == "sirepo_sandbox":
+            for u in self.hdb.sirepo_sandbox.users.keys():
+                t = T(u, self.buildt)
```
robnagler, line 23: pass in user in kwargs to T

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+        z.run_u = jc.rsconf_db.run_u
+        # Only additional config for the server is the sshd config.
+        z.run_d = systemd.docker_unit_prepare(self, jc, watch_files=[z.host.ssh_d])
+        self._network(jc, z)
+
+    def internal_build_write(self):
+        from rsconf import systemd
+
+        if self.name == "sirepo_sandbox":
+            self.append_root_bash(": nothing to do")
```
robnagler, line 47: rather ": user instances do all the installs"

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+            return
+        jc = self.j2_ctx
+        z = jc.sirepo_sandbox
+        v = []
+        x = [
+            [z.host.ssh_d, z.guest.ssh_d],
+            # SECURITY: no modifications to run_d
+            [z.run_d, z.run_d, "ro"],
```
robnagler, line 55: We also want db/src which is rw and owned by vagrant.

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+        from rsconf import systemd
+
+        if self.name == "sirepo_sandbox":
+            self.append_root_bash(": nothing to do")
+            return
+        jc = self.j2_ctx
+        z = jc.sirepo_sandbox
+        v = []
+        x = [
+            [z.host.ssh_d, z.guest.ssh_d],
```
robnagler, line 53: This should be read-only too.

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+            image=jc.sirepo_sandbox.docker_image,
+            volumes=x,
+            cmd="/usr/sbin/sshd -D -f '{}'".format(z.guest.sshd_config),
+        )
+        self.install_access(mode="700", owner=z.run_u)
+        for d in [z.host_d, z.host.ssh_d, *v]:
+            self.install_directory(d)
+        self.install_access(mode="400")
+        self.install_resource(z.host.sshd_config, jc)
+        for k, v in z.secrets.items():
```
robnagler, line 73: You need to install only the private key for the host and the public key for the user. Not both. Unlike mpi_worker, the public key for the host and the private key for ssh will reside on the user's laptop (not `~/jupyter`).

robnagler (changes requested):
We need to talk about user volumes.

rsconf/component/__init__.py
```diff
@@ -152,12 +152,12 @@ def install_perl_rpm(self, j2_ctx, rpm_base, channel=None):
...
         return r
 
-    def install_resource(self, name, j2_ctx, host_path=None):
+    def install_resource(self, name, j2_ctx, host_path=None, component_name=None):
         if not host_path:
             host_path = name
             if host_path.ext == ".sh":
                 host_path = host_path.new(ext="")
-            name = self.name + "/" + name.basename
+            name = (component_name or self.name) + "/" + name.basename
```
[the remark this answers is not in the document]
e-carlin: 566ec33

rsconf/component/sirepo_sandbox.py
```diff
@@ -0,0 +1,125 @@
...
+        if self.name == "sirepo_sandbox":
+            for u in self.hdb.sirepo_sandbox.users.keys():
+                t = T(u, self.buildt)
+                self.buildt.build_component(t)
+            return
+        self.buildt.require_component("docker", "network")
+        jc, _ = self.j2_ctx_init()
+        z = jc.sirepo_sandbox
+        z.setdefault("volumes", {})
+        # TODO(e-carlin): Better way then building the path manually
```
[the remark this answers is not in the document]
e-carlin: 566ec33

rsconf/package_data/sirepo_sandbox/sshd_config.jinja
e-carlin: 566ec33

rsconf/package_data/sirepo_sandbox/sshd_config.jinja
e-carlin: 566ec33

rsconf/package_data/sirepo_sandbox/sshd_config.jinja
e-carlin: 566ec33

rsconf/package_data/sirepo_sandbox/sshd_config.jinja
e-carlin: 566ec33

rsconf/component/sirepo_sandbox.py
e-carlin: 566ec33

rsconf/component/sirepo_sandbox.py
e-carlin: aeab41b

rsconf/component/sirepo_sandbox.py
e-carlin: aeab41b

rsconf/component/sirepo_sandbox.py
e-carlin: aeab41b

rsconf/component/sirepo_sandbox.py
e-carlin: aeab41b

rsconf/component/sirepo_sandbox.py
e-carlin: af3b1f8

rsconf/component/sirepo_sandbox.py
e-carlin: 147cc16

rsconf/component/sirepo_sandbox.py
e-carlin: 1f49f6a

rsconf/component/sirepo_sandbox.py
e-carlin: 1f49f6a

robnagler:
@e-carlin what's the status? I think we need to get this moving into production so we can start coding.

e-carlin:
I need to address the last couple of comments. I'll fix and request another review by EOD

rsconf/component/sirepo_sandbox.py
e-carlin: 8d1bb90

rsconf/component/sirepo_sandbox.py
e-carlin: 8d68e84

rsconf/component/sirepo_sandbox.py
e-carlin: They are included through "volumes" configuration (although I added rw 2eccc4d). There is a problem with src. The container exits immediately upon start with no message. My guess is it is because it is missing home-env and something calls some function from it on start. 

I need to fix https://github.com/radiasoft/rsconf/issues/224 and then can test more.

e-carlin:
@robnagler ready for another review. All problems solved except https://github.com/radiasoft/rsconf/pull/217#discussion_r936085938 which we should talk about.

rsconf/component/sirepo_sandbox.py
e-carlin: [d2604ae](https://github.com/radiasoft/rsconf/pull/217/commits/d2604ae449df7786c57f4ee0ce59e23414c19ed3)

rsconf/component/__init__.py
```diff
@@ -332,6 +329,38 @@ def _bash_append_and_dst(
...
 
+    def _gen_host_and_identity_ssh_keys(
+        self, j2_ctx, filename, visibility, password=None
+    ):
+        from rsconf import db
+
+        res = PKDict()
+        b = db.secret_path(j2_ctx, filename, visibility=visibility, directory=True)
+        res.host_key_f = b.join("host_key")
+        res.host_key_pub_f = res.host_key_f + ".pub"
```
robnagler, line 340: I think `b.new(ext="pub")` is how you are supposed to do this in py.path.

rsconf/component/__init__.py
```diff
@@ -332,6 +329,38 @@ def _bash_append_and_dst(
         self._bash_append(host_path, md5=md5)
         return dst
 
+    def _gen_host_and_identity_ssh_keys(
```
robnagler, line 332: Seems like this should be public.

rsconf/component/__init__.py
```diff
@@ -332,6 +329,38 @@ def _bash_append_and_dst(
...
+        self, j2_ctx, filename, visibility, password=None
```
robnagler, line 333: Default to ""?

rsconf/component/__init__.py
```diff
@@ -332,6 +329,38 @@ def _bash_append_and_dst(
...
+    ):
+        from rsconf import db
```
robnagler, line 335: could be imported up top

rsconf/component/devbox.py
```diff
@@ -0,0 +1,113 @@
...
+        p = db.random_string()
+        res = super()._gen_host_and_identity_ssh_keys(
+            jc, "devbox/" + self.user_name, visibility="host", password=p
+        )
+        for k in list(res.keys()):
+            if k not in ("host_key_f", "identity_pub_f"):
+                res.pkdel(k)
+        f = db.secret_path(jc, _PASSWD_SECRET_JSON_F, visibility="host")
+        o = pkjson.load_any(f) if f.check() else PKDict()
+        o[self.user_name] = p
```
robnagler, line 102: This will change the password every time. You want to do this first so you can get the password for the user or create it. See component.rsconf.host_init.

rsconf/systemd.py
```diff
@@ -71,9 +71,9 @@ def custom_unit_enable(
     unit_enable(compt, j2_ctx)
 
 
-def custom_unit_prepare(compt, j2_ctx, watch_files=()):
+def custom_unit_prepare(compt, j2_ctx, watch_files=(), run_d=None):
     """Must be first call"""
-    run_d = unit_run_d(j2_ctx, compt.name)
+    run_d = run_d or unit_run_d(j2_ctx, compt.name)
```
robnagler, line 76: Use a temporary or use `if not run_d: run_d = ...`.  I don't like assigning run_d if already assigned.

rsconf/component/devbox.py
```diff
@@ -0,0 +1,113 @@
...
+        if "user_name" not in self:
+            for u in self.hdb.devbox.users.keys():
+                t = T(u, self.buildt, user_name=u)
+                self.buildt.build_component(t)
+            return
+        self.buildt.require_component("docker", "network")
+        jc, _ = self.j2_ctx_init()
+        z = jc.devbox
+        z.setdefault("volumes", ["jupyter", "src"])
+        z.host_d = systemd.unit_run_d(jc, "devbox_" + self.user_name)
```
robnagler, line 33: seems like compt.name could be devbox_user_name and set that on create. Then everything will be consistent.

robnagler (changes requested):
Would be good to add some cases to tests. I think we might want to do something with random_string to get reliable output, but you could skip that for now

rsconf/component/__init__.py
e-carlin: I get tripped up in Python with classes and what should be public but I believe since only instances of the class call this method (using self or super) private is correct. Agree?

```python
rsconf (210-sirepo-sandbox)$ gpy _gen_host_and_identity_ssh_keys
./rsconf/component/__init__.py:332:    def _gen_host_and_identity_ssh_keys(
./rsconf/component/devbox.py:34:        z.secrets = self._gen_host_and_identity_ssh_keys(jc)
./rsconf/component/devbox.py:85:    def _gen_host_and_identity_ssh_keys(self, jc):
./rsconf/component/devbox.py:89:        res = super()._gen_host_and_identity_ssh_keys(
./rsconf/component/mpi_worker.py:26:        z.secrets = self._gen_host_and_identity_ssh_keys(jc, z)
./rsconf/component/mpi_worker.py:95:    def _gen_host_and_identity_ssh_keys(self, jc, z):
./rsconf/component/mpi_worker.py:96:        return super()._gen_host_and_identity_ssh_keys(
./rsconf/component/mpi_worker.py:130:            s = self._gen_host_and_identity_ssh_keys(jc, z)
```

rsconf/component/devbox.py
```diff
@@ -0,0 +1,113 @@
...
+                    r += f"/{c}"
+                self.append_root_bash(
+                    f"rsconf_clone_repo https://github.com{r}.git {p} {jc.rsconf_db.run_u}"
+                )
+        self.install_access(mode="400")
+        self.install_resource(z.host.sshd_config, jc)
+        for k, v in z.secrets.items():
+            self.install_abspath(v, z.host[k])
+
+    def _network(self, jc, z):
```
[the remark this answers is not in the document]
e-carlin: 0b2eeca

rsconf/component/__init__.py
e-carlin: 0b2eeca

rsconf/component/__init__.py
e-carlin: 0b2eeca

rsconf/component/__init__.py
e-carlin: 0b2eeca

rsconf/component/devbox.py
e-carlin: 0b2eeca

rsconf/component/devbox.py
e-carlin: fixing the compt.name removed the need for supplying run_d 5d2406d

e-carlin: Whoops, previous comment was on the wrong comment of yours. 

Thanks for seeing compt.name. Cleaned things up 5d2406d

rsconf/systemd.py
e-carlin: fixing the compt.name removed the need for supplying run_d https://github.com/radiasoft/rsconf/commit/5d2406d37cba0a5b455f327bc6df141c05c6010a

rsconf/component/__init__.py
robnagler: It's inconsistent with the other methods in component.T. I understand the principle of most private. The only guidance in PEP8 is: 

> Use one leading underscore only for non-public methods and instance variables.

and

> Another category of attributes are those that are part of the “subclass API” (often called “protected” in other languages). Some classes are designed to be inherited from, either to extend or modify aspects of the class’s behavior. When designing such a class, take care to make explicit decisions about which attributes are public, which are part of the subclass API, and which are truly only to be used by your base class.

Underscores do not seem to come into play here other than reading between the lines.

I think there is no such thing as "protected" in Python. In BOP we labeled methods with an "internal_" prefix to mean protected, but Perl adopts no such convention and there is no name mangling like in Python for double underscore.

In this particular case, I vote for consistency for the other functions. You certainly can add this as a /discuss issue.

rsconf/component/devbox.py
```diff
@@ -23,22 +23,21 @@ def internal_build_compile(self):
 
         if "user_name" not in self:
             for u in self.hdb.devbox.users.keys():
-                t = T(u, self.buildt, user_name=u)
-                self.buildt.build_component(t)
+                self.buildt.build_component(
+                    T(f"{self.name}_{u}", self.buildt, user_name=u)
```
robnagler, line 27: nice :)

rsconf/component/devbox.py
```diff
@@ -82,25 +81,21 @@ def internal_build_write(self):
...
     def _gen_host_and_identity_ssh_keys(self, jc):
         from rsconf import db
 
-        p = db.random_string()
+        f = db.secret_path(jc, _PASSWD_SECRET_JSON_F, visibility="host")
+        o = pkjson.load_any(f) if f.check() else PKDict()
         res = super()._gen_host_and_identity_ssh_keys(
-            jc, "devbox/" + self.user_name, visibility="host", password=p
+            jc,
+            "devbox/" + self.user_name,
```
robnagler, line 91: I liked your module_name concept. Maybe this could be part of the creation of the instance in the base class so that a component has a name and a module_name, and then overriden above when creating the child instances.

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -27,32 +27,42 @@ rsconf_append() {
...
         install -m 600 -o "$user" -g "$user" /dev/null "$keys_f"
     fi
     rsconf_edit_no_change_res=0 rsconf_append "$keys_f" "$key"
 }
 
+rsconf_clone_repo() {
+    declare repo=$1
+    declare dest=$2
+    declare user=$3
+    if [[ ! -d $dest || ! $(ls -A "$dest") ]]; then
```
robnagler, line 47: Why the need for the second check? If it is a directory, it should not be cloned.

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -2,10 +2,10 @@
 set -euo pipefail
 
 rsconf_append() {
-    local file=$1
-    local egrep=$2
+    declare file=$1
```
robnagler, line 5: Thanks.

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -27,32 +27,42 @@ rsconf_append() {
...
     if [[ ! -e $keys_f ]]; then
         install -m 600 -o "$user" -g "$user" /dev/null "$keys_f"
     fi
     rsconf_edit_no_change_res=0 rsconf_append "$keys_f" "$key"
 }
 
+rsconf_clone_repo() {
+    declare repo=$1
+    declare dest=$2
+    declare user=$3
```
robnagler, line 46: user => owner (and implicitly group). 

Perhaps we need an install umask for this case. 

Also set the umask in a subshell and run git clone  In this case we can hardwire 027 as the umask for now until we need something different.

Note the weirdness about [git clone and umask](https://stackoverflow.com/a/11335009).

robnagler (changes requested):

rsconf/component/__init__.py
e-carlin: Make sense. Thanks for the details. 367ae5c

rsconf/package_data/rsconf/rsconf.sh
e-carlin: We need to be able to handle the fact that `devbox_<user>.sh` is going to create the full directory structure. Ex
```
rsconf_install_access '700' 'vagrant' 'vagrant'
rsconf_install_directory '/srv/devbox_xxx/jupyter'
rsconf_install_directory '/srv/devbox_xxx/src'
rsconf_install_directory '/srv/devbox_xxx/src/biviosoftware'
rsconf_install_directory '/srv/devbox_xxx/src/biviosoftware/home-env'
rsconf_clone_repo https://github.com/biviosoftware/home-env.git /srv/devbox_xxx/src/biviosoftware/home-env vagrant
```

rsconf/package_data/rsconf/rsconf.sh
e-carlin: 9ffe221

Looks like that weirdness was fixed in [1.7.11.3](https://github.com/git/git/commit/45d4fdc2dc74c657a9c2e95bf04aed539fdcb0a4). We have 1.8.3.1 on centos7.

rsconf/package_data/rsconf/rsconf.sh
robnagler: OK.

rsconf/component/devbox.py
e-carlin: I'm not sure I understand what you're looking for. Is this it 9d156a0 ?

rsconf/component/mpi_worker.py
```diff
@@ -13,6 +13,11 @@
 
 
 class T(component.T):
+    def gen_host_and_identity_ssh_keys(self, jc, z):
+        return super().gen_host_and_identity_ssh_keys(
+            jc, self.name + "/" + z.user, visibility="channel"
```
robnagler, line 18: z.user could be assigned self.user_name, or devbox could put user in z. The goal is to share as much code with gen_host_and_identity_ssh_keys. I don't think this needs to be a subclassed method.

robnagler (changes requested):
I gave a bunch of comments on the commit, which means you are going to have to find them. I edited one of the comments.

rsconf/component/mpi_worker.py
e-carlin: How does this look ccdb931 ?

rsconf/component/__init__.py
```diff
@@ -80,6 +84,55 @@ def build_write(self):
         self._root_bash.extend(self._root_bash_aux)
         self.buildt.write_root_bash(self.name, self._root_bash)
 
+    def gen_host_and_identity_ssh_keys(
+        self, j2_ctx, visibility, password_filename=None
```
robnagler, line 88: I think just "encrypt_identity=False" would be better, and have the password_filename be `f"{self.module_name}_ssh_passphrase.json"`. It's not a password, and it isn't "devbox_auth". It's a passphrase for the ssh identity key, but that's long so ssh_passphrase is clear, I think.

rsconf/component/__init__.py
```diff
@@ -80,6 +84,55 @@ def build_write(self):
...
+    ):
+        @contextlib.contextmanager
```
robnagler, line 90: Too complicated and unnecessary work. You can write the passphrase before the generate happens. Logic should be:

```py
for f:
    if not f.exists():
        process_call(
        ...
       _pass() if encrypt_identity and f == res.identity_f else "",
```
_pass would be simpler:

```py    
def _pass():
     s = db.secret_path(j2_ctx, f"{self.module_name}_ssh_passphrase.json", visibility=visibility)
     o = pkjson.load_any(s) if s.exists() else PKDict()
     p = o.get(self.user_name)
     if p is None:
         o[self.user_name] = p = db.random_string()
         pkjson.dump_pretty(o, filename=s)
     return p
```

rsconf/component/__init__.py
```diff
@@ -80,6 +84,55 @@ def build_write(self):
...
+
+        with _persist_password(password_filename) as w:
+            res = PKDict()
+            b = db.secret_path(
+                j2_ctx,
+                f"{self.module_name}/{self.user_name}",
+                visibility=visibility,
+                directory=True,
+            )
+            res.host_key_f = b.join("host_key")
```
robnagler, line 111: Could be simplified:

```py
for x in "identity", "host_key":
    # possibly make this a function
    y = tuple((f"{x}{s}"_f", b.join(x).new(ext=s[1:]) for s in "", "_pub")
    res.update(y)
    if y[0][1].exists:
        continue
```

rsconf/component/__init__.py
```diff
@@ -157,7 +210,7 @@ def install_resource(self, name, j2_ctx, host_path=None):
             host_path = name
             if host_path.ext == ".sh":
                 host_path = host_path.new(ext="")
-            name = self.name + "/" + name.basename
+            name = (self.get("module_name") or self.name) + "/" + name.basename
```
robnagler, line 213: I think there should always be a module_name. For backward compatibility, you will need to only do this if "/" isn't already in name iwc the caller already added module_name (typical case). This is a highly constrained problem, because we never have a "name" without a "/" (you can assert this), that is, all resources are at least two levels deep. Therefore I think this trick will be sufficient to fix all cases.

robnagler (changes requested):

rsconf/component/devbox.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
+"""Sirepo development in a docker container on a remote machine
+
+:copyright: Copyright (c) 2019 RadiaSoft LLC.  All Rights Reserved.
```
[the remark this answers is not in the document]
e-carlin: 5cf71b2

rsconf/component/devbox.py
```diff
@@ -0,0 +1,104 @@
...
+"""
+from pykern.pkcollections import PKDict
+from pykern.pkdebug import pkdp
+from rsconf import component
+
+_PASSWD_SECRET_JSON_F = "devbox_auth.json"
+
+
+class T(component.T):
+    def gen_host_and_identity_ssh_keys(self, jc):
```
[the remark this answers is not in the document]
e-carlin: 5cf71b2

rsconf/component/devbox.py
```diff
@@ -0,0 +1,104 @@
...
+        res = super().gen_host_and_identity_ssh_keys(
+            jc, "host", password_filename=_PASSWD_SECRET_JSON_F
+        )
+        for k in list(res.keys()):
```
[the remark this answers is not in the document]
e-carlin: 5cf71b2

rsconf/component/__init__.py
e-carlin: 5cf71b2

rsconf/component/__init__.py
e-carlin: Much better, thanks 5cf71b2

rsconf/component/__init__.py
e-carlin: > you will need to only do this if "/" isn't already in name

I'm not sure what you mean by "this"?

I think just using self.module_name works 749c4cb. We default module_name to name. That I believe satisfies our backwards compat needs. In the devbox case module_name is devbox (not devbox_<username> which is what self.name is) so it works in that case as well.

rsconf/component/__init__.py
e-carlin: 5cf71b2

robnagler (approved):
Not the day to push to production so I'll let you merge on Monday.

robnagler:
Tests are failing so please fix before merge.

PR #232, opened 2022-08-11
fix radiasoft/rsscorpius#87: jupyter/bashrc with default config

e-carlin:


rsconf/package_data/devbox/jupyter_bashrc.jinja
```diff
@@ -0,0 +1,9 @@
+#! /bin/bash
```
e-carlin, line 1: How should we handle the user editing this file and us overwriting it? One thought is that we could add a line like `source ~/jupyter/post_jupyter_bashrc` to this file. Users could then modify their env with that file and leave this one alone.

PR #233, opened 2022-08-11
Create default jupyter bashrc

e-carlin:
- Supply sirepo ports and urls so they don't conflict with other devbox
users.
- Allow setting of package_path and sim_types

### DIFF
--- a/rsconf/component/__init__.py
+++ b/rsconf/component/__init__.py
@@ -11,7 +11,6 @@ from pykern import pkjson
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdp, pkdc, pkdlog
 from rsconf import db
-import contextlib
 import hashlib
 import re
 import subprocess
@@ -162,14 +161,6 @@ class T(PKDict):
                 **self._install_access
             ),
         )
-
-    @contextlib.contextmanager
-    def install_access_temp(self, **kwargs):
-        p = self._install_access.copy()
-        self.install_access(**kwargs)
-        yield
-        self._install_access = p
-        self.install_access(**self._install_access)
 
     def install_directory(self, host_path):
         assert (
--- a/rsconf/component/__init__.py
+++ b/rsconf/component/__init__.py
@@ -211,6 +211,7 @@ class T(PKDict):
             host_path,
             file_contents=self._render_resource(name, j2_ctx),
         )
+        return host_path
 
     def install_secret_path(
         self, filename, host_path, gen_secret=None, visibility=None
--- a/rsconf/component/__init__.py
+++ b/rsconf/component/__init__.py
@@ -11,6 +11,7 @@ from pykern import pkjson
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdp, pkdc, pkdlog
 from rsconf import db
+import contextlib
 import hashlib
 import re
 import subprocess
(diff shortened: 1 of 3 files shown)


rsconf/component/devbox.py
```diff
@@ -96,3 +102,5 @@ def _network(self, jc, z):
         z.ip = n.unchecked_public_ip()
         z.ssh_port = jc.devbox.users[self.user_name]
         n.add_public_tcp_ports([str(z.ssh_port)])
+        z.flask_port = z.ssh_port + 1000
```
robnagler, line 105: Why use a different name than SIREPO_PKCLI_SERVICE_PORT? Indeed, it might be easier to generate the export of the vars and have them appended one at a time to jupyter_bashrc.

I think the 1000 should be configurable, just as a matter of principle.

rsconf/component/__init__.py
```diff
@@ -162,6 +163,14 @@ def install_access(self, mode=None, owner=None, group=None):
             ),
         )
 
+    @contextlib.contextmanager
+    def install_access_temp(self, **kwargs):
```
robnagler, line 167: I think "temp" is not a great name. Rather, if you want this feature, just add "mode" to install_resource and have it do the temporary setting of the access.

robnagler (dismissed):

rsconf/component/__init__.py
e-carlin: I agree on the name. I left a review on this code asking for your opinion but forgot to hit submit. Anyways, I was thinking if you were ok with this pattern then I would try to find a way to make install_resource work like open. That is, it could be called as a "normal" function or as a contexmanager. In the normal function case it would do what it does now. In the contextmanager case it would do what this function is doing. Would you prefer that or adding mode to install_resource?

robnagler: install_access is designed to support "grouping" easily in the install process. The idea is that most installs have a common pattern: resources are read only, directories are writable, etc. I don't think there is a need for context. In this specific case, if you move the [`install_access`](https://github.com/radiasoft/rsconf/pull/233/files#diff-b972143365867703eab8e161882070dc75a75c0547d7f5744cc8281cd7dc7fa8R63) inside the for loop, you don't need a temporary change to the install_access.

rsconf/component/devbox.py
e-carlin: Configurable port 9032a86.

If you mean why did I not name it z.service_port I thought the name "service" was too generic and didn't give the reader any indication for what the port is used for. If you mean why did I not name it z.SIREPO_PKCLI_SERVICE_PORT I didn't consider that one. That looks kind of odd to me but is that what you're looking for?

Appending them to jupyter_bashrc sounds like a good idea. Is there a feature in rsconf for appending to a file like that?  Or should I add one (maybe to rsconf.sh)? Or should I build up the string in python and the render in jinja? Or...?

rsconf/component/__init__.py
e-carlin: 9032a86

rsconf/component/devbox.py
robnagler: z.service_port is in this context understandable, because this is a configurator and it is configuring something that defines that name, no matter how bad it is. The word "flask" doesn't appear anywhere else in the configuration so it is confusing if you are just looking at the configurator component.

robnagler: > Is there a feature in rsconf for appending to a file like that? 

self.rsconf_append

e-carlin: https://github.com/radiasoft/rsconf/pull/233/commits/b7d412c9f148cace74a35bd348d4d5437bd03294

PR #239, opened 2022-08-15
ELK status PR

Benbenbenin0:


rsconf/component/elasticsearch.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
+u"""create elasticsearch configuration
+:copyright: NA
+:license: NA
+"""
+import os
+from __future__ import absolute_import, division, print_function
+from pykern.pkcollections import PKDict
```
robnagler, line 8: Remove `__future__`, and from the module you copied it from. Add 

rsconf/component/elasticsearch.py
```diff
@@ -0,0 +1,78 @@
...
+        # move elastic.repo file into /etc/yum.repos.d
+        path = Path('/etc/yum.repos.d/elastic.repo')
+        if path.is_file():
+            self.append_root_bash('echo Elastic Yum Repo: Found by Elasticsearch')
+        else: 
+            self.append_root_bash('echo Elastic Yum Repo: Moving by Elasticsearch')
+            self.append_root_bash('mv /home/vagrant/src/radiasoft/rsconf/rsconf/package_data/elasticsearch/elastic.repo /etc/yum.repos.d')
+            
+        # clear yum cache and ready metadata for needed repos, pointless step ??
+        self.append_root_bash('sudo yum clean all')
```
robnagler, line 51: Too much code. The infrastructure handles this for you.

rsconf/component/elasticsearch.py
```diff
@@ -0,0 +1,78 @@
...
+        self.append_root_bash('echo "Elasticsearch touched"')
+        
+        # install/update java jdk
+        self.append_root_bash('rsconf_yum_install java-1.8.0-openjdk')
+        
+        # download + install public signing key
+        self.append_root_bash('sudo rpm --import https://packages.elastic.co/GPG-KEY-elasticsearch')
+        
+        # move elastic.repo file into /etc/yum.repos.d
+        path = Path('/etc/yum.repos.d/elastic.repo')
```
robnagler, line 43: rsconf does this for you.

rsconf/component/elasticsearch.py
```diff
@@ -0,0 +1,78 @@
...
+        self.buildt.require_component('base_all') # componants generally need base_all
+
+        # print to test component is being touched
+        self.append_root_bash('echo "Elasticsearch touched"')
+        
+        # install/update java jdk
+        self.append_root_bash('rsconf_yum_install java-1.8.0-openjdk')
+        
+        # download + install public signing key
+        self.append_root_bash('sudo rpm --import https://packages.elastic.co/GPG-KEY-elasticsearch')
```
robnagler, line 40: rsconf runs as root so no sudo. Importing a key is something you only do once as well as adding the repo. We should add something to rsconf.sh like `rsconf_yum_import_key`. The repo file itself would be installed as a normal file (install_resource would work).

rsconf/component/elasticsearch.py
```diff
@@ -0,0 +1,78 @@
...
+class T(component.T):
+
+    def internal_build(self): # Difference between internal_build and internal_build_compile ??
+        self.buildt.require_component('base_all') # componants generally need base_all
+
+        # print to test component is being touched
+        self.append_root_bash('echo "Elasticsearch touched"')
+        
+        # install/update java jdk
+        self.append_root_bash('rsconf_yum_install java-1.8.0-openjdk')
```
robnagler, line 37: We would add this to download/installers/redhat-base. I think this is fine for now.

rsconf/component/elasticsearch.py
```diff
@@ -0,0 +1,78 @@
...
+        else: 
+            self.append_root_bash('echo Elastic Yum Repo: Moving by Elasticsearch')
+            self.append_root_bash('mv /home/vagrant/src/radiasoft/rsconf/rsconf/package_data/elasticsearch/elastic.repo /etc/yum.repos.d')
+            
+        # clear yum cache and ready metadata for needed repos, pointless step ??
+        self.append_root_bash('sudo yum clean all')
+        self.append_root_bash('sudo yum makecache')
+        
+        # yum install
+        self.append_root_bash('rsconf_yum_install elasticsearch')
```
robnagler, line 55: Look at rsconf/systemd.py to help with this. You are installing elasticsearch as a service. It will have a run directory, etc. that we define. The unit file needs to be properly configured, and rsconf.systemd takes care of that, too.

robnagler (changes requested):
We should pair on this. There are some important concepts that have been missed.

robnagler:
Won't be able to merge so close for now. Can always pull out later.

PR #242, opened 2022-08-15
Fix #234: Stop users from getting kicked out of the ssh session on error

### LINKED ISSUE #234: devbox: kicked out of ssh session on error
try `echo 'exit 1' > x.sh && bash x.sh`. You'll be dropped back onto whatever machine you ssh'ed in from

e-carlin:
The set -e causes errors to result in exit when a command resulted in an error.


rsconf/component/devbox.py
```diff
@@ -98,23 +95,32 @@ def _gen_secrets(self, jc):
...
+                raise ValueError(
+                    f"single quote in value={v} of bash variable name={name}"
+                )
+            self.rsconf_append(
+                path,
+                f"export {name}='{v}'",
+            )
+
+        self.install_access(mode="600")
+        self.install_ensure_file_exists(path)
```
e-carlin, line 112: I believe this causes the file to be appended to instead of overwritten. Is that desired?

rsconf/component/__init__.py
```diff
@@ -451,6 +458,11 @@ def _assert_host_path(host_path):
     )
 
 
+def _bash_quote(value):
+    v = str(value).replace("\\", r"\\").replace("'", r"\'")
+    return f"$'{v}'"
```
e-carlin, line 463: Neat. For anyone else or my future self the `$''` syntax is called ansi-c quoting https://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html

rsconf/component/devbox.py
```diff
@@ -98,23 +95,32 @@ def _gen_secrets(self, jc):
...
+            if "'" in v:
+                raise ValueError(
+                    f"single quote in value={v} of bash variable name={name}"
+                )
+            self.rsconf_append(
+                path,
+                f"export {name}='{v}'",
+            )
+
+        self.install_access(mode="600")
```
e-carlin, line 111: How do you decide these? Just least privilege possible? I looked on my machine and saw 644 so I went with it...

rsconf/component/devbox.py
robnagler: Correct. We want to allow local modifications.

rsconf/component/devbox.py
robnagler: Nothing like this should be world readable. If it is installed from open source and read only, yes, that's fine, and necessary for NERSC.

PR #248, opened 2022-08-23
Fix #246 added `X11Forwarding yes`

### LINKED ISSUE #246: devbox add X11Forwarding yes to sshd config

### LINKED ISSUE #247: devbox sshd log to stderr

robnagler:
Fix #247 added `-e` to sshd cmd


robnagler:
One thing to note about the `-e` is that this message is always output: `Attempt to write login records by non-root user (aborting)`. However, it doesn't mean "aborting sshd". It means it is not going to write /var/log/btmp, etc. It's a very old message, and people get confused by it (include me just now).

PR #251, opened 2022-08-25
Fix #250: add ability to restrict ports

### LINKED ISSUE #250: add ability to restrict ports to certain ip ranges
Like
```
network:
   restrict_ports:
       https: [ 192.168.1.1/24 ]
```

e-carlin:
You can define
```
network:
  restrict_ports:
    https: [ 192.168.1.1/24 ]
```

That will restrict traffic on https port to traffic from only 192.168.1.1/24.

http will be added for you.

### DIFF
--- a/rsconf/component/network.py
+++ b/rsconf/component/network.py
@@ -43,6 +43,7 @@ class T(component.T):
         self.j2_ctx = self.hdb.j2_ctx_copy()
         jc = self.j2_ctx
         z = jc.network
+        self._check_restricted_ports(z)
         z.trusted_nets = tuple(sorted(z.trusted.keys()))
         z.pksetdefault(
             blocked_ips=[],
@@ -150,6 +151,19 @@ class T(component.T):
             for w in check:
                 assert not p in z[w], "port {} already in {}".format(p, w)
             z[which].append(p)
+
+    def _check_restricted_ports(self, z):
+        if "restrict_ports" not in z:
+            return
+        assert (
+            "443" not in z.restrict_ports
+        ), f"use https instead of 443 in restrict_ports={z.restrict_ports}"
+        assert (
+            "http" not in z.restrict_ports and "80" not in z.restrict_ports
+        ), f"only define https not http or 80 in restrict_ports={z.restrict_ports}"
+        if "https" not in z.restrict_ports:
+            return
+        z.restrict_ports["http"] = z.restrict_ports.https
 
     def _defroute(self, routes):
         defroute = None
--- a/rsconf/component/network.py
+++ b/rsconf/component/network.py
@@ -100,6 +100,8 @@ class T(rsconf.component.T):
         jc = self.j2_ctx
         jc = self.j2_ctx
         z = jc.network
+        # Order matters: _restricted_public_tcp_ports modifed public_tcp_ports
+        # to remove ports that that in the restricted set.
         self._restricted_public_tcp_ports(z)
         if not "_devs" in z:
             # no devices, no network config
--- a/rsconf/component/network.py
+++ b/rsconf/component/network.py
@@ -1,19 +1,18 @@
-# -*- coding: utf-8 -*-
 """create network config
 
 We disable NetworkManager, because we are managing the network now.
 It's unnecessary complexity. NetworkManager (NM) create
 ``network-scripts`` files.
 
-:copyright: Copyright (c) 2018 RadiaSoft LLC.  All Rights Reserved.
+:copyright: Copyright (c) 2018-2022 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
-from __future__ import absolute_import, division, print_function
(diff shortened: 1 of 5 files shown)


e-carlin:
I just realized `iptables-restore v1.4.21: ! not allowed with multiple source or destination IP addresses` . Do we need to support a list of ips for the restrict or can it just be a single ip address CIDR range for now?

e-carlin:
nvm it needs to be a list. fixing...

e-carlin:
@robnagler ready for a review

robnagler:
@e-carlin we can talk about a lot of changes. Review my changes.

tests/pkcli/build_data/1-fconf.out/srv/host/v9.radia.run/etc/sysconfig/iptables
```diff
@@ -6,14 +6,12 @@
...
 -A INPUT -p icmp -j ACCEPT
-    -A INPUT -i em2 -p tcp -m state --state NEW -m tcp -s 192.168.1.1/24,127.0.0.1 --dport https -j ACCEPT
-    -A INPUT -i em2 -p tcp -m state --state NEW -m tcp --dport https -j DROP
-    -A INPUT -i em2 -p tcp -m state --state NEW -m tcp -s 192.168.1.1/24,127.0.0.1 --dport http -j ACCEPT
-    -A INPUT -i em2 -p tcp -m state --state NEW -m tcp --dport http -j DROP
     -A INPUT -i lo -j ACCEPT
     -A INPUT -i em1 -j ACCEPT
-    -A INPUT -i em2 -p tcp -m state --state NEW -m tcp --match multiport --dports 3100,3101,9999,http,https,pop3s,smtp,submission -j ACCEPT
 -A INPUT -i em2 -m state --state RELATED,ESTABLISHED -j ACCEPT
+    -A INPUT -i em2 -p tcp -m state --state NEW -m tcp --match multiport --dports 3100,3101,9999,http,pop3s,smtp,submission -j ACCEPT
```
e-carlin, line 12: This still has http. Are we assuming nginx always redirects http to https?

rsconf/component/network.py
```diff
@@ -100,6 +100,7 @@ def internal_build_write(self):
         jc = self.j2_ctx
         jc = self.j2_ctx
         z = jc.network
+        self._restricted_public_tcp_ports(z)
```
e-carlin, line 103: I think it is worth commenting this bit of code that it is important it is the last thing to mutate the lists of ports.

e-carlin (commented):
Looks good. 

I still don't understand drop vs reject. You say that drop tells the attacker something. [This](https://serverfault.com/questions/157375/reject-vs-drop-when-using-iptables#:~:text=The%20difference%20is%20that%20the,need%20to%20wait%20for%20timeout.&text=DROP%20will%20cause%20packet%20to,cannot%20hide%20existence%20of%20server.) leads me to believe that reject tells them something and drop tells them nothing.

tests/pkcli/build_data/1-fconf.out/srv/host/v9.radia.run/etc/sysconfig/iptables
robnagler: We don't allow anything over http except a redirect to https. For this case, we only need https blocked. This is simpler and clearer for now. Eventually we could support port ranges, but it's an unnecessary complication now.

robnagler:
> I still don't understand drop vs reject. You say that drop tells the attacker something. This leads me to believe that reject tells them something and drop tells them nothing.

If you read the all the replies, you will see the debate, e.g. [this comment](https://serverfault.com/questions/157375/reject-vs-drop-when-using-iptables#comment1418172_157376):

> DROP will cause packet to timeout but it cannot hide existence of server. Try nmap -sS targethost sometime. It will be able to tell the difference between true closed and filtered ports based on SYN behavior.

You have to DROP everything on the host for it to be "invisible". If the host is in any way visible (ping response, for example), then a DROP tells you more than a REJECT.

rsconf/component/network.py
robnagler: c24d8fa added comments to iptables as well.

e-carlin:
> If you read the all the replies, you will see the debate, e.g. [this comment](https://serverfault.com/questions/157375/reject-vs-drop-when-using-iptables#comment1418172_157376)

Interesting, thanks.

This comment further down was also illuminating:
> Without iptables rules, a closed port defaults to REJECT. If you DROP every port it's a fine alternative (your host appears down) but if you DROP only specific ports you're actually giving them more information than REJECT. If someone uses a blanket probe like nmap, specific ports that drop packets will appear as "filtered", which means they know you have a service there that you're hiding.

robnagler:
That's the way I think about it. 

Anything that is on the Internet is visible in some way. Without that, you are doing to spend a lot of time debugging. If you care about these things, you use a firewall (also discussed in the thread).

PR #265, opened 2022-10-14
Fix #263: create mongodb component (for raydata)

### LINKED ISSUE #263: create mongodb component

e-carlin:
000.yml:
```yml
mongod:
  port: 27018
  ip: 127.0.0.1
  database:
    name: databroker
```

secret/000.yml:
```yml
mongod:
  admin:
    username: admin
    password: password123
  database:
    username: databroker
    password: passwordabc
```

### DIFF
--- a/rsconf/component/mongod.py
+++ b/rsconf/component/mongod.py
@@ -27,7 +27,7 @@ class T(component.T):
         z.data_d = z.run_d.join(_DATA_DIR)
         z.log_d = z.run_d.join(_LOG_DIR)
         z.pid_d = pkio.py_path("/run/mongod")
-        z.pid_file_path = z.pid_d.join("mongod.pid")
+        z.pid_f = z.pid_d.join("mongod.pid")
         z.conf_f = z.run_d.join(_MONGO_CONF_F)
 
     def internal_build_write(self):
--- a/rsconf/component/mongod.py
+++ b/rsconf/component/mongod.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+"""create nginx configuration
+
+:copyright: Copyright (c) 2017 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from pykern.pkcollections import PKDict
+from pykern.pkdebug import pkdp
+from rsconf import component
+
+_DATA_DIR = "db"
+
+_MONGO_CONF_F = "/etc/mongod.conf"
+
+
+class T(component.T):
+    def internal_build_compile(self):
+        self.buildt.require_component("base_all")
+        self.j2_ctx, _ = self.j2_ctx_init()
+
+    def internal_build_write(self):
+        from rsconf import systemd
+
+        z = self.j2_ctx.mongod
+        z.run_u = "mongod"
+        d = self.j2_ctx.rsconf_db.host_run_d.join(self.name)
+        z.data_d = d.join(_DATA_DIR)
+        self.install_access(mode="700", owner=z.run_u)
+        self.install_directory(d)
+        self.install_directory(z.data_d)
+        self.install_access(mode="400", owner=z.run_u)
+        self.install_resource(
+            self.name + "/mongod.conf",
+            self.j2_ctx,
+            host_path=_MONGO_CONF_F,
+        )
+        self.append_root_bash_with_main(self.j2_ctx)
+        systemd.unit_enable(self, self.j2_ctx)
+        self.service_prepare([_MONGO_CONF_F])
+        self.rsconf_service_restart()
--- a/rsconf/component/mongod.py
+++ b/rsconf/component/mongod.py
@@ -22,7 +22,6 @@ class T(component.T):
 
         self.buildt.require_component("base_all")
         self.j2_ctx, z = self.j2_ctx_init()
-        z.run_d = systemd.custom_unit_prepare(self, self.j2_ctx)
         z.run_u = "mongod"
         z.data_d = z.run_d.join(_DATA_DIR)
         z.log_d = z.run_d.join(_LOG_DIR)
--- a/rsconf/component/mongod.py
+++ b/rsconf/component/mongod.py
@@ -34,7 +34,6 @@ class T(component.T):
 
         z = self.j2_ctx.mongod
         z.run_u = "mongod"
-        z.run_d = self.j2_ctx.rsconf_db.host_run_d.join(self.name)
         z.data_d = z.run_d.join(_DATA_DIR)
         z.log_d = z.run_d.join(_LOG_DIR)
         z.pid_d = pkio.py_path("/run/mongod")
--- a/rsconf/component/mongod.py
+++ b/rsconf/component/mongod.py
@@ -40,7 +40,7 @@ class T(component.T):
             self.install_directory(z[f"{d}"])
         self.install_access(mode="400", owner=z.run_u)
         self.install_resource(
-            self.name + f"/{_MONGO_CONF_F}",
+            f"{self.name}/{_MONGO_CONF_F}",
             self.j2_ctx,
             host_path=z.conf_f,
         )
--- a/rsconf/component/mongod.py
+++ b/rsconf/component/mongod.py
@@ -22,23 +22,19 @@ class T(component.T):
 
         self.buildt.require_component("base_all")
         self.j2_ctx, z = self.j2_ctx_init()
-        # TODO(e-carlin): this is a goofy catch 22
-        z.mongo_conf_path = self.j2_ctx.rsconf_db.host_run_d.join(self.name).join(
-            _MONGO_CONF_F
-        )
-        z.run_d = systemd.custom_unit_prepare(self, self.j2_ctx, (z.mongo_conf_path,))
+        z.run_d = systemd.custom_unit_prepare(self, self.j2_ctx)
+        z.run_u = "mongod"
+        z.data_d = z.run_d.join(_DATA_DIR)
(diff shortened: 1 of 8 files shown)


rsconf/package_data/mongod/mongod.conf.jinja
```diff
@@ -0,0 +1,19 @@
...
+security:
+  authorization: enabled
+storage:
+  dbPath: {{ mongod.data_d }}
+  journal:
+    enabled: true
+systemLog:
+  destination: file
+  logAppend: true
+  path: /var/log/mongodb/mongod.log
```
robnagler, line 19: Does mongod come with a logrotate? If not, we should build our own and put the logs in /srv/mongod.

rsconf/package_data/mongod/mongod.conf.jinja
```diff
@@ -0,0 +1,19 @@
+# for documentation of all options, see:
+#   http://docs.mongodb.org/manual/reference/configuration-options/
+net:
+  port: {{ mongod.port }}
+  bindIp: {{ mongod.ip }}
+processManagement:
+  fork: true
+  pidFilePath: /var/run/mongodb/mongod.pid
```
robnagler, line 8: I think this should be /run/mongod/mongod.pid. I have a feeling we need to create /run/mongod, since mongod runs as mongod. Maybe it is created in the rpm? We should use consistent naming (everything mongod) even they don't. /var/run is outdated.

rsconf/component/mongod.py
```diff
@@ -0,0 +1,40 @@
...
+
+_DATA_DIR = "db"
+
+_MONGO_CONF_F = "/etc/mongod.conf"
+
+
+class T(component.T):
+    def internal_build_compile(self):
+        self.buildt.require_component("base_all")
+        self.j2_ctx, _ = self.j2_ctx_init()
```
robnagler, line 19: This should have the definitions and a unit_prepare, which returns run_d which is saved in j2_ctx in that call. This will call compt.service_prepare, which should be before the enable. run_u, data_d, should be set here.

rsconf/component/mongod.py
```diff
@@ -0,0 +1,40 @@
...
+:copyright: Copyright (c) 2017 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from pykern.pkcollections import PKDict
+from pykern.pkdebug import pkdp
+from rsconf import component
+
+_DATA_DIR = "db"
+
+_MONGO_CONF_F = "/etc/mongod.conf"
```
robnagler, line 13: It would be ideal to move this /srv/mongod

robnagler (changes requested):
Please add to the build test so we know the files are created correctly.

rsconf/component/mongod.py
e-carlin: 6c04a19

rsconf/component/mongod.py
e-carlin: 6c04a19

rsconf/package_data/mongod/mongod.conf.jinja
e-carlin: 6c04a19

/var/run/... is created by the rpm. Just setting pidFilePath to /run/... doesn't work because pidFile in the systemd unit file from the rpm is hard coded to /var/run/... . I think I've resolved all of it by just using a custom unit.

rsconf/package_data/mongod/mongod.conf.jinja
e-carlin: It doesn't. Added one 6c04a19

rsconf/package_data/mongod/main.sh.jinja
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+
+mongod_main() {
+    declare v=6.0
+    declare f=/etc/yum.repos.d/mongodb-org-"$v".repo
+    if [[ -e $f ]]; then
+        install_info "repo file $f exists, mongo already installed"
```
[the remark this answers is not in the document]
e-carlin: 6c04a19

rsconf/component/mongod.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class T(component.T):
+    def internal_build_compile(self):
+        from rsconf import systemd
+
+        self.buildt.require_component("base_all")
+        self.j2_ctx, z = self.j2_ctx_init()
+        # TODO(e-carlin): this is a goofy catch 22
+        z.mongo_conf_path = self.j2_ctx.rsconf_db.host_run_d.join(self.name).join(
```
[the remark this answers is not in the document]
robnagler: unit_prepare automatically adds run_d and the systemd files:     compt.service_prepare([f, d] + list(watch_files))

rsconf/package_data/mongod/mongod.conf.jinja
robnagler: I think that's better.

rsconf/package_data/mongod/logrotate.conf.jinja
```diff
@@ -0,0 +1,10 @@
+{{ mongod.log_d }}/*.log
+/var/log/mongodb/mongod.log {
+    missingok
+    notifempty
+    postrotate
+        kill -SIGUSR1 $(cat {{ mongod.pid_file_path }}) &> /dev/null
```
robnagler, line 6: incorrect (funny enough github flagged this):
Screen Shot 2022-10-17 at 11 00 47 AM

rsconf/component/mongod.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def internal_build_write(self):
+        from rsconf import systemd
+        from rsconf.component import logrotate
+
+        z = self.j2_ctx.mongod
+        z.run_u = "mongod"
+        z.data_d = z.run_d.join(_DATA_DIR)
+        z.log_d = z.run_d.join(_LOG_DIR)
+        z.pid_d = pkio.py_path("/run/mongod")
```
robnagler, line 39: I don't see pid_d used except on the next line. Could just have a constant (PID_FILE), which would be consistent with the above.

rsconf/component/mongod.py
```diff
@@ -0,0 +1,55 @@
...
+        z.pid_file_path = z.pid_d.join("mongod.pid")
```
robnagler, line 40: You call it `ip`, not `bind_ip`, but you call this `pid_file_path`. I realize you are trying to keep consistent names. I fee like we have a model here where we name things `_f` and should be consistent with that as opposed to be consistent with the files. Not sure, just an opinion.

rsconf/component/mongod.py
```diff
@@ -0,0 +1,55 @@
...
+        z.mongo_conf_path = z.run_d.join(_MONGO_CONF_F)
+        self.install_access(mode="700", owner=z.run_u)
+        for d in "run_d", "data_d", "log_d", "pid_d":
+            self.install_directory(z[f"{d}"])
+        self.install_access(mode="400", owner=z.run_u)
+        self.install_resource(
+            self.name + f"/{_MONGO_CONF_F}",
```
robnagler, line 47: why is there a "+" here and not all in the `f`.

rsconf/component/mongod.py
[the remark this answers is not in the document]
robnagler: z.conf_f would be a better name.

robnagler (changes requested):

rsconf/package_data/mongod/mongod.conf.jinja
e-carlin: They do. That code is outdated after  https://github.com/radiasoft/rsconf/commit/6c04a196e62e07639a1c84a428ab1cd8fdfebe21

See
https://github.com/radiasoft/rsconf/commit/6c04a196e62e07639a1c84a428ab1cd8fdfebe21#diff-a74861d485a09fe805e46df8af0f6b164c63eddb08c2685b38593e0204cd10acR19

rsconf/component/mongod.py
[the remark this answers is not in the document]
e-carlin: Doh, thanks.

I struggle with file vs path. Sometimes file means basename. Other times it means full path. 

0ba9277

rsconf/package_data/mongod/logrotate.conf.jinja
e-carlin: Can you elaborate on incorrect? I think what I wrote redirects stdout and stderr to /dev/null which I believe is what I want.

https://www.gnu.org/software/bash/manual/html_node/Redirections.html#:~:text=and%20standard%20error%3A-,%26%3Eword,-and

rsconf/component/mongod.py
e-carlin: No reason. 8ea280f

rsconf/package_data/mongod/logrotate.conf.jinja
robnagler: Well, you learn something everyday: https://superuser.com/a/1650179

I'm not sure I buy the reasoning, e.g. if you don't space things right you get different results, which is probably a more common issue:

```sh
$ echo & > x
[1] 20361

[1]+  Done                    echo
$ echo > & x
bash: syntax error near unexpected token `&'
```

In the first case, you wrote a zero length file and the command executed in background. In the second, it failed.

Anyway, it's fine to keep it. I won't change my reasoning.

I don't know why github flagged the & other than it being confused as to the file type.

robnagler (approved):

robnagler:
You merge.

PR #269, opened 2022-10-24
fix #268: install mongo first so mongod user is available

### LINKED ISSUE #268: mongod: chown: invalid user: ‘mongod’
rsc errored out with `chown: invalid user: ‘mongod’`. 

The problem is [mongod/main.sh.jinja](https://github.com/radiasoft/rsconf/blob/caeeffb5f678e06735d84ef59d85cf148232b269/rsconf/package_data/mongod/main.sh.jinja#L4) installs the mongo rpm which creates a mongod user. But, [main is appended](https://github.com/radiasoft/rsconf/blob/caeeffb5f678e06735d84ef59d85cf148232b269/rsconf/component/mongod.py#L48) after we try to install files [with the mongod user](https://github.com/radiasoft/rsconf/blob/caeeffb5f678e06735d84ef59d85cf148232b269/rsconf/component/mongod.py#L38). So, we need to move the main appending to earlier in the install so the mongod user is available.

e-carlin:


e-carlin:
merged to keep moving but please review.

rsconf/component/mongod.py
```diff
@@ -33,6 +33,7 @@ def internal_build_write(self):
         from rsconf.component import logrotate
 
         z = self.j2_ctx.mongod
+        self.append_root_bash_with_main(self.j2_ctx)
```
robnagler, line 36: There are two parts to main: installing the rpm and configuring mongod. You should probably split these up. You can append_root_bash_with_resource to install the rpm so that the file installs work and then append with another resource that configures mongod (without the last restart which seems unnecessary).  The configuration needs a sentinel that a user is created, but that probably can be checked via mongosh.

Also the `<<EOF` should be `<<'EOF`` so that bash doesn't accidentally evaluate a password with a dollar sign or whatever.

PR #271, opened 2022-10-24
fix #270: merged dicts so config from 000.yml is used too

### LINKED ISSUE #270: raydata_scan_monitor: AssertionError: raydata.pkcli.scan_monitor.catalog_names: config value missing and is required
https://github.com/radiasoft/rsconf/blob/494c6887214eae718dcce240854833f3d1f7e51e/rsconf/component/raydata_scan_monitor.py doesn't merge z (which has configuration) with z.db_d

e-carlin:


e-carlin:
Again, merged to keep moving but please review. I'm unsure about the copy.

rsconf/component/raydata_scan_monitor.py
```diff
@@ -49,5 +50,8 @@ def internal_build_write(self):
 
     def _unit_env(self, z):
         return pkconfig.to_environ(
-            ["raydata.pkcli.*"], values={"raydata.pkcli.scan_monitor.db_dir": z.db_d}
+            ["raydata.pkcli.*"],
+            values=PKDict(copy.deepcopy(z)).pkupdate(
```
robnagler, line 54: The "z"  is [already be a copy](https://github.com/radiasoft/rsconf/pull/271/files#diff-f1d8f33eea3bde336b215690e4d2ef52f73ef109cd8a3ded3178f1be28b000c8L22) so there should be no need for a further copy. I'm not sure why this particular environment variable not be written in internal_build_compile.

PR #127, opened 2020-10-14
WIP: Issue/126 sirepo jupyterhub

e-carlin:
Most of the work for running jupyterhub under Sirepo. I left some TODO's around creating dependencies between sirepo, sirepo_jupyterhub, and jupyterhub so the right volumes are mounted and accessible in the right containers (ex so we can migrate data to new jupyterhub). I'm not exactly sure how we want to handle this and how to implement. May be best to discuss in real-time. 

Getting everything running on a remote server using ssh LocalForward required some hacks. There were issues with re-directs dropping the port number and jupyterhub blocking requests due to them being cross origin. In the end I setup a reverse proxy on my mac and edited /etc/hosts to make it all work. My nginx foo is weak so maybe there is a simpler way to get it working that doesn't require a reverse proxy.

~/.ss/config on mac:
```
Host <redacted>
  HostName <redacted>
  Port  <redacted>
  User <redcated>
  LocalForward 8080 sirepo.v4.radia.run:443
```

nginx on mac:
```sh
brew install nginx
cd /usr/local/etc/nginx
sudo openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout cert.key -out crt.pem # for fqdn use sirepo.v4.radia.run
cat <<EOF > /usr/local/etc/nginx/nginx.conf
worker_processes  1;
error_log /usr/local/var/log/nginx/error.log debug;

events {
    worker_connections  1024;
}

http {
    include       mime.types;
    default_type  application/octet-stream;

    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for"';

    sendfile        on;
    keepalive_timeout  65;

    proxy_buffer_size   128k;
    proxy_buffers   4 256k;
    proxy_busy_buffers_size   256k;
    server {
      listen 443 ssl;
      server_name sirepo.v4.radia.run;
      ssl_certificate      cert.pem;
      ssl_certificate_key  cert.key;
      proxy_http_version 1.1;
      proxy_next_upstream error timeout invalid_header http_500 http_502 http_503 http_504;
      proxy_read_timeout 600s;
      proxy_send_timeout 600s;
      proxy_connect_timeout 600s;
      proxy_redirect off;
      proxy_set_header Host $host:$server_port;
      proxy_set_header Proxy "";
      proxy_set_header Via $remote_addr;
      proxy_set_header X-Real-Ip $remote_addr;
[...]

PR #275, opened 2022-10-27
fix #127: make logrotate scirpt portable to /bin/sh

e-carlin:


rsconf/package_data/mongod/logrotate.conf.jinja
```diff
@@ -2,7 +2,7 @@
     missingok
     notifempty
     postrotate
-        kill -SIGUSR1 $(cat {{ systemd.pid_file }}) &> /dev/null
```
robnagler, line 5: This is non-traditional. /bin/kill supports `-USR1` which is how I've always seen it.

e-carlin: thanks, c86507f

The `|| true` doesn't strike me as very fail fast. It would've masked the current error we're having right?

robnagler: Yes, it isn't fail fast but we probably don't want to cause a failure in this case. I'm not sure if logrotate continues after errors. Better to buy into the way everybody else does it for this particular problem. I wouldn't write logrotate this way at all, but it's the way it works...

robnagler (approved):

PR #279, opened 2022-11-28
fix #147: allow specifying arbitrary data to include in daemon_aux

### LINKED ISSUE #147: do_indent() got an unexpected keyword argument 'indentfirst'
https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.indent

indentfirst was renamed to indent

e-carlin:


rsconf/package_data/docker/daemon.json.jinja
```diff
@@ -19,4 +19,7 @@
     "storage-opts": [
         "overlay2.override_kernel_check=true"
     ]
+    {% for k in docker.get('daemon_aux', [])  %}
```
robnagler, line 22: Invalid file, missing common on line above. I think it would actually be ideal to generate daemon.json without jinja when I see this. Just merge PKDicts and call dump_json.

robnagler (changes requested):

rsconf/package_data/docker/daemon.json.jinja
e-carlin: Nice catch. Thanks.

ae5db64

rsconf/component/docker.py
```diff
@@ -83,6 +78,33 @@ def internal_build(self):
             run_d=z.run_d,
         )
 
+    def _install_daemon_json(self, z):
+        p = self.tmp_path()
+        d = PKDict(
```
robnagler, line 83: Since we are on the subject of copies, it's much better to not use the copy and do an update, or better:

```py
PKDict({
    "hosts": ...
    ...
    "data-root": ...
}).update(z.get("daemon_aux", {}))
```

While it's not a performance issue per se, these types of "habits" are not good to have in production code.

robnagler (approved):
Comment about style but not important for merge.

rsconf/component/docker.py
e-carlin: 1bd52a8

e-carlin:
@robnagler the test data changed. Please take a look.

robnagler (approved):

PR #281, opened 2022-11-29
fix #280: fix tests to account for leaving dst

### LINKED ISSUE #280: tests failing on master
This is blocking my pr https://github.com/radiasoft/rsconf/actions/runs/3569650206/jobs/5999811224

I can reproduce on my machine. According to GH tests passed for the most recent commit (or at least that is what I think the green checkmark next to the commit means). Not sure what's going on...

e-carlin:


robnagler (commented):
lgtm

PR #296, opened 2023-02-23
Fix #295: custom image for a devbox user

### LINKED ISSUE #295: devobx: image per user
Add the ability to configure a specific image for a user. This will aid in f36 deployment where we want users to be able to test devbox while other users continue to work on f32 devbox.

e-carlin:


rsconf/component/devbox.py
```diff
@@ -53,10 +53,11 @@ def internal_build_write(self):
                 [z.run_d, z.run_d, "ro"],
             ],
         )
+        u = jc.devbox.users[self.user_name]
         systemd.docker_unit_enable(
             self,
             jc,
-            image=jc.devbox.docker_image,
+            image=u.docker_image if isinstance(u, dict) else jc.devbox.docker_image,
```
robnagler, line 60: This pattern repeats. Compile is for handling this: converting arbitrary input formats into canonical form with defaults filled in. Put self.docker_image, defaulting to devbox.docker_image.

tests/pkcli/build_data/1.in/db/000.yml
```diff
@@ -446,6 +446,9 @@ host:
         users:
           devtech3: 3100
           fullstackdude: 3101
+          user_w_custom_image:
```
robnagler, line 449: just call the user `custom_image`. Prefixing with user is unnecessary and perhaps confusing.

robnagler (changes requested):

rsconf/component/devbox.py
e-carlin: d20cddb

tests/pkcli/build_data/1.in/db/000.yml
e-carlin: d20cddb

rsconf/component/devbox.py
```diff
@@ -35,6 +35,12 @@ def internal_build_compile(self):
         z.run_u = jc.rsconf_db.run_u
         # Only additional config for the server is the sshd config.
         z.run_d = systemd.custom_unit_prepare(self, jc, watch_files=[z.host.ssh_d])
+        u = jc.devbox.users[self.user_name]
+        self._docker_image = jc.devbox.docker_image
+        z.ssh_port = u
+        if isinstance(u, dict):
+            self._docker_image = u.docker_image
```
robnagler, line 42: Don't assume anything exists in the dict. Also testing for dict is problematic with dot form (u.docker_image). Test for PKDict. Always use get. There's self and z. I think you could just rely on z as follows:

```py
if isinstance(u, PKDict):
    z.docker_image = u.get('docker_image', z.docker_image)
    z.ssh_port = u.ssh_port
else:
    z.ssh_port  = u
```

Another point to make is that bop uses self.name to get configuration that's specific, e.g. jc.petshop.port. You could consider just looking for `devbox_<user>.docker_image`. I'm not sure that's better, but it would be consistent with bop, which does this similarly.

rsconf/component/devbox.py
```diff
@@ -56,7 +62,7 @@ def internal_build_write(self):
         systemd.docker_unit_enable(
             self,
             jc,
-            image=jc.devbox.docker_image,
+            image=self._docker_image,
```
robnagler, line 65: This would be z.

robnagler (changes requested):

rsconf/component/devbox.py
e-carlin: 2d2b6a6

rsconf/component/devbox.py
e-carlin: 2d2b6a6

robnagler (approved):

PR #310, opened 2023-04-14
Fix #309: Allow specification of a minimum docker version

### LINKED ISSUE #309: Allow specification of docker version
Either pegged or a min version.

e-carlin:
If the mimimum version is not met Docker will be updated.

### DIFF
--- a/rsconf/component/docker.py
+++ b/rsconf/component/docker.py
@@ -32,7 +32,7 @@ class T(component.T):
         self.buildt.require_component("base_all", "db_bkp")
         j2_ctx = self.hdb.j2_ctx_copy()
         z = j2_ctx.docker
-        self._check_min_version(z)
+        self._check_min_software_version(z)
         systemd.unit_prepare(
             self,
             j2_ctx,
@@ -79,12 +79,13 @@ class T(component.T):
             run_d=z.run_d,
         )
 
-    def _check_min_version(self, z):
-        z.setdefault("min_version", "0.0.0")
-        p = [int(v) for v in z.min_version.split(".")]
+    def _check_min_software_version(self, z):
+        z.setdefault("min_software_version", "0.0.0")
+        p = [int(v) for v in z.min_software_version.split(".")]
         if len(p) != 3:
-            raise ValueError(f"min_version={z.min_version} needs to be in X.X.X format")
-        z.min_version = p
+            raise ValueError(
+                f"min_software_version={z.min_software_version} needs to be in X.X.X format"
+            )
 
     def _install_daemon_json(self, z):
         p = self.tmp_path()
--- a/rsconf/component/docker.py
+++ b/rsconf/component/docker.py
@@ -32,6 +32,7 @@ class T(component.T):
         self.buildt.require_component("base_all", "db_bkp")
         j2_ctx = self.hdb.j2_ctx_copy()
         z = j2_ctx.docker
+        self._check_min_version(z)
         systemd.unit_prepare(
             self,
             j2_ctx,
@@ -77,6 +78,13 @@ class T(component.T):
             run_u=z.run_u,
             run_d=z.run_d,
         )
+
+    def _check_min_version(self, z):
+        z.setdefault("min_version", "0.0.0")
+        p = [int(v) for v in z.min_version.split(".")]
+        if len(p) != 3:
+            raise ValueError(f"min_version={z.min_version} needs to be in X.X.X format")
+        z.min_version = p
 
     def _install_daemon_json(self, z):
         p = self.tmp_path()
--- a/rsconf/package_data/dev/db/000.yml.jinja
+++ b/rsconf/package_data/dev/db/000.yml.jinja
@@ -81,7 +81,7 @@ default:
     on_calendar: "6:30"
   docker:
     iptables: false
-    min_version: "23.0.0"
+    min_software_version: "23.0.0"
   docker_cache:
     host: ''
   docker_registry:
--- a/rsconf/package_data/dev/db/000.yml.jinja
+++ b/rsconf/package_data/dev/db/000.yml.jinja
@@ -81,6 +81,7 @@ default:
     on_calendar: "6:30"
   docker:
     iptables: false
+    min_version: "23.0.0"
   docker_cache:
     host: ''
   docker_registry:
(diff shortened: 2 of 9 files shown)


rsconf/package_data/dev/db/000.yml.jinja
```diff
@@ -81,6 +81,7 @@ default:
     on_calendar: "6:30"
   docker:
     iptables: false
+    min_version: 23.0.0
```
robnagler, line 84: min_version is not descriptive enough. min_software_version is clearer.

rsconf/package_data/docker/main.sh.jinja
```diff
@@ -28,23 +40,45 @@ Then re-run rsconf
...
 
 docker_main() {
     docker_install
 }
+
+docker_should_update() {
+    docker_should_update_helper Client || docker_should_update_helper Server
+}
+
+docker_should_update_helper() {
```
robnagler, line 70: We haven't use "helper" in the past. I would usually call this docker_should_update_one or docker_should_update_do or even docker_should_update_component.

rsconf/package_data/docker/main.sh.jinja
```diff
@@ -28,23 +40,45 @@ Then re-run rsconf
...
+    declare docker_type="$1"
+    declare s="$(docker version --format="{{ '{{ json . }}' }}" | jq --raw-output ".$docker_type.Version")"
```
robnagler, line 72: docker_type is component: https://docs.docker.com/engine/reference/commandline/version/
outside quotes are unnecessary and obscure the other quotes.

rsconf/package_data/docker/main.sh.jinja
```diff
@@ -28,23 +40,45 @@ Then re-run rsconf
...
 docker_main() {
     docker_install
 }
+
+docker_should_update() {
+    docker_should_update_helper Client || docker_should_update_helper Server
+}
+
+docker_should_update_helper() {
+    declare docker_type="$1"
```
robnagler, line 71: quotes unnecessary

rsconf/package_data/docker/main.sh.jinja
```diff
@@ -28,23 +40,45 @@ Then re-run rsconf
...
+    declare s="$(docker version --format="{{ '{{ json . }}' }}" | jq --raw-output ".$docker_type.Version")"
+    declare -a v=( ${s//./ } )
+    declare -a d=( {{ docker.min_version[0] }} {{ docker.min_version[1] }} {{ docker.min_version[2] }} )
```
robnagler, line 74: might be better to have a string here like "docker._version_parts" instead of doing the separating in jinja. I use underscore for keys that shouldn't be set by a user.

rsconf/package_data/docker/main.sh.jinja
```diff
@@ -28,23 +40,45 @@ Then re-run rsconf
...
+    for i in "${!d[@]}"; do
+        if (( "${v[i]}" > "${d[i]}" )); then
```
robnagler, line 76: quotes and ${} are unnecessary. This works in bash:`((  v[i] > d[i] ))`. This only works for arithmetic conditionals. 

You could also consider creating a function like this

```sh
docker_version_num() {
    local str=$1
    printf '%d%06d%06d' ${str//./}
}
```

This then could be used like this:

```sh
docker_version_lt() {
    local left=$1
    local right=$2
    (( $(docker_version_num "$left") < $(docker_version_num "$right") ))
}
```

This makes the assumption that the numbers are integers, which is also made by the loop implementation. And, it assumes that minor and patch versions are less than 1M, which is reasonable, I feel.

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -572,8 +572,15 @@ rsconf_yum_install() {
         cmd=reinstall
     fi
     if (( ${#todo[@]} > 0 )); then
-        yum "$cmd" --color=never -y -q "${todo[@]}"
+        rsconf_yum "$cmd" "${todo[@]}"
     fi
 }
 
+
+rsconf_yum() {
```
robnagler, line 580: I think `install_yum` does what you want.

rsconf/component/docker.py
```diff
@@ -78,6 +79,13 @@ def internal_build(self):
             run_d=z.run_d,
         )
 
+    def _check_min_version(self, z):
+        z.setdefault("min_version", "0.0.0")
+        p = [int(v) for v in z.min_version.split(".")]
+        if len(p) != 3:
+            raise ValueError(f"min_version={z.min_version} needs to be in X.X.X format")
+        z.min_version = p
```
robnagler, line 87: This is not good practice. Do not overwrite a configuration value with a different type. See my discussion below about `_version_parts`.

robnagler (changes requested):

rsconf/package_data/dev/db/000.yml.jinja
e-carlin: 0349609

rsconf/package_data/docker/main.sh.jinja
e-carlin: 0349609

rsconf/package_data/docker/main.sh.jinja
e-carlin: 0349609

rsconf/package_data/docker/main.sh.jinja
e-carlin: 0349609

rsconf/package_data/docker/main.sh.jinja
e-carlin: 0349609

rsconf/package_data/docker/main.sh.jinja
e-carlin: 0349609

rsconf/component/docker.py
e-carlin: 0349609

rsconf/package_data/rsconf/rsconf.sh
e-carlin: da85b31

robnagler (approved):

PR #332, opened 2023-05-19
fix #323: raydata: create notebook_d

### LINKED ISSUE #323: raydata: create notebook_d
The raydata service now requires a mounted directory that contains the notebook zips. This will work in much the same way as the intake directory.

rorour:


rsconf/component/raydata_scan_monitor.py
```diff
@@ -29,6 +31,7 @@ def internal_build_compile(self):
         )
         z.db_d = z._run_d.join(_DB_SUBDIR)
         z.intake_d = z._run_d.join(_INTAKE_D)
+        z.notebook_d = z._run_d.join(_NOTEBOOK_D)
```
e-carlin, line 34: need to add this to jc.sirepo so it is populated in the env file.

e-carlin (changes requested):

rsconf/component/raydata_scan_monitor.py
rorour: 950e53473a38140aab077fbfeab13fe54c366e2c

e-carlin (approved):

PR #360, opened 2023-07-20
Fix #352 add least_conn to sirepo proxy backend scheduling

### LINKED ISSUE #352: sirepo tornado nginx route least_conn
The default is round robin, which is fine for short lived requests. With web sockets, `least_conn` is better.

robnagler:


rorour:
This doesn't seem related to your changes, but every time I try to run `rsc` it ends with

```
curl -L -s -S -n http://v3.radia.run:2916/biviosoftware/container-perl/radiasoft-download.sh
radiasoft-download.sh: no #! at start of file: /var/tmp/radia-run-21414-4486/20230720180155-radiasoft-download.sh
If you don't know what to do, please contact support@radiasoft.net.
```

Do you know what would cause this?

e-carlin:
Try running the curl command above that error (`curl -L -s -S -n http://v3.radia.run:2916/biviosoftware/container-perl/radiasoft-download.sh`) and see what happens. My guess is you will get a 404 back. If that's true then the probable cause is that v3 isn't serving the files. Probably because the webserver isn't running. You can start it up with `bash run/nginx/start.sh` ([from the development notes](https://github.com/radiasoft/rsconf/wiki/Development)). Obviously all of those are guesses as to what's going on so if that fix doesn't work lmk.

rorour:
@e-carlin Thanks. apparently I was just missing biviosoftware/container-perl on v3. Works now.

rorour (approved):
Looks good to me

PR #371, opened 2023-08-24
Fix #370: standardize rsiviz port configuration and enable devbox for rsiviz

### LINKED ISSUE #370: rsiviz networking + devboxes
It would be easier if we ran through nginx in development. Then we wouldn't have to change the client notebook (won't need to remove secret or change https->http). In addition, it will keep us behind the secret which is safer.

We should allow a range of ports to be used by people in development. All of them should be proxied through nginx.

Maybe have some scheme like 7001->8001 so we can easily know the ports the services should be running on. 

In addition, switch to host networking. Our default iptables configuration is to reject the entire FORWARD chain which blocks docker traffic when doing port mapping.

e-carlin:


rsconf/component/rsiviz.py
```diff
@@ -19,27 +19,33 @@ def internal_build_compile(self):
         jc, z = self.j2_ctx_init()
         z._run_u = jc.rsconf_db.run_u
         self.j2_ctx_pksetdefault(
-            PKDict(pykern=PKDict(pkconfig=PKDict(channel=jc.rsconf_db.channel)))
+            PKDict(
+                pykern=PKDict(
```
robnagler, line 23: I think this is problematic. Rather, it should be cascaded explicitly, e.g. rsiviz.server_port gets assigned to .pykern.pkasyncio.server_port. Otherwise, you can only have one pkasyncio service per host. I would also probably define rsiviz.index_iframe_port and cascade that to rsiviz.pkcli.service....

rsconf/component/rsiviz.py
```diff
@@ -19,27 +19,33 @@ def internal_build_compile(self):
...
     def internal_build_write(self):
-        from rsconf.component import db_bkp
         from rsconf.component import nginx
-        from rsconf.component import docker
 
         jc = self.j2_ctx
         z = jc[self.name]
         d = self.__run_d.join(_DB_SUBDIR)
         jc.nginx.pkupdate(
             index_port=8880,
```
robnagler, line 45: This would be good as constants. You could just have a pkdict that is passed to pkupdate.

tests/pkcli/build_data/1.in/db/000.yml
```diff
@@ -460,6 +460,15 @@ host:
           custom_image:
             docker_image: radiasoft/custom-image
             ssh_port: 3102
+          rsivizdevbox:
```
robnagler, line 463: I was a little confused by this. Maybe rsivizcoder?

rsconf/component/devbox.py
```diff
@@ -108,34 +119,47 @@ def _gen_secrets(self, jc):
...
+            path,
+        )
+        self._env(
+            "RSIVIZ_PKCLI_SERVICE_DICE_NETWORK_CLUSTER_INTERFACE_ADDRESS",
+            ipaddress.ip_address(u.rsiviz.ip_base) + 1,
+            path,
+        )
+        self._env(
+            "RSIVIZ_PKCLI_SERVICE_DICE_NETWORK_NETWORK_DISCOVERY_ADDRESS",
+            f"{u.rsiviz.ip_base}:5555",
```
robnagler, line 160: Better to have this as a contant.

rsconf/component/devbox.py
```diff
@@ -108,34 +119,47 @@ def _gen_secrets(self, jc):
...
+        if not isinstance(u, PKDict) or not "rsiviz" in u:
+            return
+        self._env(
+            "PYKERN_PKASYNCIO_SERVER_PORT",
+            u.rsiviz.port_base,
+            path,
+        )
+        self._env(
+            "RSIVIZ_PKCLI_SERVICE_DICE_NETWORK_CLUSTER_INTERFACE_ADDRESS",
+            ipaddress.ip_address(u.rsiviz.ip_base) + 1,
```
robnagler, line 155: This is really confusing, but I guess it works...

robnagler (changes requested):

rsconf/component/devbox.py
e-carlin: Any suggestions on how to make it less confusing? We just need to be on different networks and this seemed like a simply way to accomplish that.

rsconf/component/rsiviz.py
e-carlin: > Otherwise, you can only have one pkasyncio service per host

The value is set in /srv/rsiviz/env so another service could set a different port in its /srv/foo/env. We could only have one rsiviz service per host (because their ports would conflict) but that doesn't seem very problematic because, to my knowledge, we don't have a way of running multiple rsiviz services on a host.

But, cascade makes sense ff62898

rsconf/component/rsiviz.py
e-carlin: ff62898

tests/pkcli/build_data/1.in/db/000.yml
e-carlin: ff62898

rsconf/component/devbox.py
e-carlin: ff62898

rsconf/component/devbox.py
robnagler: I meant the fact that `ip_address() + 1` does something (not sure what). I don't like overloading in this way. It is going to add 1 to the ip address. Not that confusing I guess.

rsconf/component/rsiviz.py
robnagler: It shouldn't be j2_ctx_pksetdefault. This should be a pkudpate. The problem is that if someone already has set any of these pykern values in the yaml, it will not override with the values being set here.

robnagler (changes requested):
The pksetdefault needs to be changed.

rsconf/component/rsiviz.py
e-carlin: 3cba232

robnagler (approved):

PR #380, opened 2023-09-04
Fix #375: rsiviz expose service ports and enable tls

### LINKED ISSUE #375: rsiviz expose nginx ports and enable tls
For rsiviz we have nginx sitting in front of the service. We need make the nginx ports we are listening to publicly available (iptables) and enable tls

e-carlin:


rsconf/component/rsiviz.py
```diff
@@ -45,6 +45,9 @@ def internal_build_compile(self):
             # is there a way to just use it and not specify cmd?
             docker_exec=f"bash {db.user_home_path(z._run_u)}/.radia-run/start",
         )
+        self.buildt.get_component("network").add_public_tcp_ports(
```
robnagler, line 48: add_public_tcp_ports should convert to str, not here.

rsconf/package_data/rsiviz/nginx.conf.jinja
```diff
@@ -1,17 +1,21 @@
 client_max_body_size 200m;
```
robnagler, line 1: I don't think this should be at this scope.

robnagler (changes requested):

rsconf/package_data/rsiviz/nginx.conf.jinja
e-carlin: 38327fc

rsconf/component/rsiviz.py
e-carlin: 38327fc

rsconf/component/network.py
```diff
@@ -142,14 +142,15 @@ def unchecked_public_ip(self):
         return None
 
     def _add_ports(self, which, ports):
-        assert len(ports[0]) > 2, "invalid ports: {}".format(ports)
+        l = [str(p) for p in ports]
```
robnagler, line 145: The temporary is unnecessary. You could leave the assertion the way it was, because it is clearer (testing arguments before touching them).

robnagler (dismissed):

rsconf/component/network.py
e-carlin: Is this what you mean be55952?

robnagler: The code actually makes no sense. What it was trying to fix was someone passing a correct port at a time when we weren't passing int ports. Here's where it was added: https://github.com/radiasoft/rsconf/commit/5d25bebff7230ebc2f1c305009d8c2135c68e07d

I see now why you had converted it to a str for a test. However, that test was for the name of the port ("domain", "smtp", etc.).  It would fail incorrectly if the port was 80 or 25 (int). Therefore, the assert should be removed or replaced with something a lot more intelligent: int ports need to be > 0 and str ports need to be found in /etc/services. That would be a nice assertion, but it's really not going to work well.

Furthermore, the assertion later doesn't check for collisions between 443 and https. The most correct software would convert all named ports to ints, and look for collisions.

Not worth implementing right now.

e-carlin: Make sense. Thanks for the explanation.

e-carlin:
@robnagler ok to merge?

robnagler:
Yes

PR #386, opened 2023-09-15
Fix #385: Support for perl rpm rollbacks

### LINKED ISSUE #385: reinstall rolled back perl rpms
In rsconf_install_perl_rpm
```bash
    elif rsconf_is_perl_rpm_rollback "$rpm_base" "$rpm_version"; then
        reinstall=1
    fi
```

schellj:
Not very familiar with this repo or versed in bash programming, so there may well be better ways to do this :)

### DIFF
--- a/rsconf/package_data/rsconf/rsconf.sh
+++ b/rsconf/package_data/rsconf/rsconf.sh
@@ -610,7 +610,7 @@ rsconf_yum_install() {
     done
     declare cmd="${rsconf_yum_install_cmd:-install}"
     if [[ ! $cmd =~ ^((re)?install|downgrade)$ ]]; then
-        install_err "$cmd: unexpected yum command"
+        install_err "unexpected value rsconf_yum_install_cmd=$cmd"
     fi
     if (( ${#todo[@]} > 0 )); then
         yum "$cmd" --color=never -y -q "${todo[@]}"
--- a/rsconf/package_data/rsconf/rsconf.sh
+++ b/rsconf/package_data/rsconf/rsconf.sh
@@ -313,21 +313,16 @@ rsconf_install_symlink() {
 }
 
 rsconf_is_perl_rpm_rollback() {
-    declare base_rpm=$1
-    declare previous_rpm=$2
-    declare this_rpm=$3
-    declare versions=()
-    for rpm in $previous_rpm $this_rpm; do
-        if [[ $rpm =~ ^$base_rpm\-([[:digit:]]{8}\.[[:digit:]]{6}-[[:digit:]])\. ]]; then
-            versions+=( ${BASH_REMATCH[1]} )
-        else
-            return 1
-        fi
-    done
-    if [[ ${versions[0]} > ${versions[1]} ]]; then
-        return 0
+    declare prev_version=$1
+    declare install_version=$2
+    if ! rsconf_perl_rpm_version prev_version "$prev_version"; then
+        return 1
     fi
-    return 1
+    rsconf_perl_rpm_version install_version "$install_version"
+    if [[ ! ${install_version:-} ]]; then
+        install_err "internal error with rsconf_perl_rpm_version parsing (install_version=$2)"
+    fi
+    [[ $prev_version > $install_version ]]
 }
 
 rsconf_main() {
@@ -369,6 +364,18 @@ rsconf_mkdir() {
     # Create the directory (and parents) with strict permissions;
     # Subsequent permissions will be created after
     install -d -o root -g root -m 700 "$d"
+}
+
+rsconf_perl_rpm_version() {
+    declare var=$1
+    declare rpm=$2
+    if [[ ! $rpm =~ -([[:digit:]]{8}\.[[:digit:]]{6}) ]]; then
+        if [[ $rpm =~ not.installed ]]; then
+            return 1
+        fi
+        install_err "invalid version format rpm=$rpm"
+    fi
+    eval "$var='${BASH_REMATCH[1]}'"
 }
 
 rsconf_radia_run_as_user() {
--- a/rsconf/package_data/rsconf/rsconf.sh
+++ b/rsconf/package_data/rsconf/rsconf.sh
@@ -264,13 +264,16 @@ rsconf_install_perl_rpm() {
     # both fresh install and update, which install does, but
     # it doesn't return an error if the update isn't done.
     # You just have to check so this way is more robust
-    declare reinstall=
+    declare install_cmd=
     if [[ $rpm_version == $prev_version ]]; then
         if rpm --verify "$rpm_base"; then
             return
         fi
         install_info "$rpm_version: rpm is modified, reinstalling"
-        reinstall=1
+        install_cmd=reinstall
+    elif rsconf_is_perl_rpm_rollback "$rpm_base" "$prev_version" "$rpm_version"; then
+        install_info "$rpm_version older than $prev_version, downgrading"
+        install_cmd=downgrade
     fi
     declare tmp=$rpm_file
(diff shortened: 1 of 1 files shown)


rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -587,9 +608,9 @@ rsconf_yum_install() {
             todo+=( "$x" )
         fi
     done
-    declare cmd=install
-    if [[ ${rsconf_yum_reinstall:-} ]]; then
-        cmd=reinstall
+    declare cmd="${rsconf_yum_install_cmd:-install}"
+    if [[ ! $cmd =~ ^((re)?install|downgrade)$ ]]; then
```
robnagler, line 612: In general, we don't write code like this, but it does make the code clearer. I guess I'm just lazy. :)

For the error messages (below), we have been using "<msg> <name>=<val>" format. The traditional shell "<val>: <msg>" is suboptimal when there are multiple values with errors. In this case include `rsconf_yum_install_cmd` so we know what the problem is, e.g.  `unexpected value rsconf_yum_install_cmd=$cmd`.

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -264,13 +264,16 @@ rsconf_install_perl_rpm() {
     # both fresh install and update, which install does, but
     # it doesn't return an error if the update isn't done.
     # You just have to check so this way is more robust
-    declare reinstall=
+    declare install_cmd=
```
robnagler, line 267: Nice change. This is a local variable so we tend to use single letters. (I know this is not always the case in rsconf.sh, but it's older than the single letter variable policy.) This function is long so it's probably fine to leave as is. Bash is harder, because function syntax in bash is annoying.

robnagler (changes requested):

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -309,6 +312,24 @@ rsconf_install_symlink() {
     rsconf_service_file_changed "$new"
 }
 
+rsconf_is_perl_rpm_rollback() {
+    declare base_rpm=$1
+    declare previous_rpm=$2
+    declare this_rpm=$3
+    declare versions=()
+    for rpm in $previous_rpm $this_rpm; do
+        if [[ $rpm =~ ^$base_rpm\-([[:digit:]]{8}\.[[:digit:]]{6}-[[:digit:]])\. ]]; then
```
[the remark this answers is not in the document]
schellj: So, the caller of `rsconf_perl_rpm_version` would have to check the return value and throw it's own `install_err`, wouldn't it? Not sure it's worth pushing the version parsing into a function in that case.

Also, I wasn't throwing an error if the pattern didn't match because if this is the first time the rpm is being installed, `previous_rpm` would be `package $base is not installed`. I suppose it might be better to check for that.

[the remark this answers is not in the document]
robnagler: The rollback case only happens if it is already installed. That should be checked, because it will be confusing. I think the code should look like:

```bash
rsconf_is_perl_rpm_rollback() {
    declare prev_version=$1
    declare install_version=$2
    if ! rsconf_perl_rpm_version prev_version "$prev_version"; then
        return 1
    fi
    rsconf_perl_rpm_version install_version "$install_version"
    if [[  ! ${install_version:-} ]]; then
        install_error "internal error with rsconf_perl_rpm_version parsing (install_version=$2)"
    fi
    [[ $prev_version > $install_version ]]
}

rsconf_perl_rpm_version() { 
    declare var=$1
    declare rpm=$2
    if ! [[ $rpm =~ -([[:digit:]]{8}\.[[:digit:]]{6}-[[:digit:]])\. ]]; then
        if [[ $rpm =~ not.installed ]]; then
            return 1
        fi
        install_err "invalid version format rpm=$rpm"
    fi
    eval "$var='${BASE_REMATCH[1]}'"
}
```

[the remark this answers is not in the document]
schellj: b70d5af

rsconf/package_data/rsconf/rsconf.sh
schellj: f44cd01

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -309,6 +312,24 @@ rsconf_install_symlink() {
...
+            versions+=( ${BASH_REMATCH[1]} )
+        else
+            return 1
+        fi
+    done
+    if [[ ${versions[0]} > ${versions[1]} ]]; then
```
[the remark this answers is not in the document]
schellj: b70d5af

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -309,6 +312,24 @@ rsconf_install_symlink() {
...
+rsconf_is_perl_rpm_rollback() {
+    declare base_rpm=$1
+    declare previous_rpm=$2
+    declare this_rpm=$3
+    declare versions=()
+    for rpm in $previous_rpm $this_rpm; do
+        if [[ $rpm =~ ^$base_rpm\-([[:digit:]]{8}\.[[:digit:]]{6}-[[:digit:]])\. ]]; then
+            versions+=( ${BASH_REMATCH[1]} )
+        else
+            return 1
```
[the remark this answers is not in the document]
schellj: b70d5af

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -309,6 +312,24 @@ rsconf_install_symlink() {
     rsconf_service_file_changed "$new"
 }
 
+rsconf_is_perl_rpm_rollback() {
+    declare base_rpm=$1
+    declare previous_rpm=$2
+    declare this_rpm=$3
+    declare versions=()
```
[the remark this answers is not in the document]
schellj: Noted, but obviated with current commits.

rsconf/package_data/rsconf/rsconf.sh
schellj: I'll leave this as there are a few other locals that would need to be updated, and it's a longer function as you say.

robnagler: It's a tricky thing with shell functions, since vars are also dynamically scoped...

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -309,6 +312,19 @@ rsconf_install_symlink() {
     rsconf_service_file_changed "$new"
 }
 
+rsconf_is_perl_rpm_rollback() {
+    declare prev_version=$1
+    declare install_version=$2
+    if ! rsconf_perl_rpm_version prev_version "$prev_version"; then
+        return 1
+    fi
+    rsconf_perl_rpm_version install_version "$install_version"
```
robnagler, line 321: It would be an error to have the value of `install_version` to be "not installed". In this case, it should be:

```bash
if ! rsconf_perl_rpm_version ...; then
   install_err "<program error> install_version=$install_version"
fi
```

The problem is if `rsconf_perl_rpm_version` returns false (1), the program will just exist without really telling you anything. Unlike Python (or Perl, the way we program it), bash doesn't give stack traces.

Think of something appropriate for `<program error>`. It might just be program error.

robnagler (changes requested):

rsconf/package_data/rsconf/rsconf.sh
schellj: fd1bc24

Good catch. I couldn't think of anything better than `program error`.

robnagler (approved):

PR #394, opened 2023-10-11
Fix #284 Add rsaccounting component

### LINKED ISSUE #284: Add rsaccounting component

robnagler:
- added rsconf_install_yum_url to allow direct downloads to work right


robnagler:
@Benbenbenin0 please review today if you can

tests/pkcli/build_data/1.in/db/000.yml
```diff
@@ -520,6 +520,12 @@ host:
         # if you do not set this, you'll get to test the SLD matching:
         # fnl1i.bivio.biz: reverse dns for 216.17.132.33 does not match SLD of v5.radia.run
         myhostname: 'v4.radia.run'
+      rsaccounting:
+        docker_image: radiasoft/rsaccounting
```
Benbenbenin0, line 524: Realizing I'm unclear on how we're setting up containers for rsaccounting.

Did we initially build the docker container on the webserver host or the master host? And was the container uploaded to a registry after for future bootstraps? Or was that where build.sh came in? Sorry didn't take good notes for this part.

Benbenbenin0 (approved):

tests/pkcli/build_data/1.in/db/000.yml
robnagler: We built on the test server to avoid having to involve the registry so we could get end-to-end working.

The word "bootstrap" is incorrect in the above context. We upload to the registry to as a part of the build process so the release pipeline hosts can download the image appropriately. 

We need to support a private image so we can include in our normal staging pipeline. This will be a bit of work along with properly setting up the build and release promotion process.

PR #404, opened 2023-11-08
Fix #403: sirepo_job_supervisor nginx conf needs to allow global-resources route

### LINKED ISSUE #403: sirepo_job_supervisor nginx conf needs to allow global-resources route

e-carlin:


rsconf/package_data/sirepo_job_supervisor/nginx.conf.jinja
```diff
@@ -8,7 +8,7 @@ server {
         {# no external server-supervisor requests #}
         return 404;
     }
-    location ~ ^/(job|sim)- {
+    location ~ ^/{{ "(job|sim" + ("|global" if sirepo.feature_config.get('enable_global_resources', False) else "") + ")" }}- {
```
robnagler, line 11: This is overkill. We don't need this type of protection. We should have a fixed prefix before all URLs so that we can do what we want (avoid not founds flooding logs).

Issue #357, opened 2023-07-07
add domain expiry and validity check


robnagler:
If a domain expires, it doesn't go away right away (sometimes not for years). Check for expiry of the domain and then fail a build which either is corrected by renewing a domain or by removing it.

PR #408, opened 2023-11-13
Fix #407: More deployment fixes for viz3d sirepo app

### LINKED ISSUE #407: rsiviz deployment fixes
- [x] nginx needs to listen on public host (currently listens on private)
- [x] fix index_fmt . Need to add VIZ3D to the env name
- [x]  add public ports that will be exposed to IndeX to allowed origin in IndeX config
- [x] add trailing slash to location blocks in nginx so paths are proxied through

e-carlin:
- Nginx needs to listen on public host (currently listens on backnet)
- Add trailing / in nginx config so routes are proxied through
- Add VIZ3D to index_fmt
- Add all public ports that IndeX could be assigned to the list of allowed_origins


e-carlin:
@robnagler let me merge once you've approved. It is helpful to me to merge and then run on the build server so I can see the config diff.

rsconf/component/rsiviz.py
```diff
@@ -1,26 +1,44 @@
...
 
 
 class T(component.T):
     def internal_build_compile(self):
         self.buildt.require_component("nginx")
         jc, z = self.j2_ctx_init()
-        self.__host = self.hdb.rsconf_db.host
+
+        n = self.buildt.get_component("network")
+        if pkconfig.in_dev_mode():
```
e-carlin, line 21: @robnagler I don't really have a clue what I'm doing in this if/else block. What I want accomplished is for the generated nginx config to look something like
```
server {
    listen <self.__host>:12002;
    server_name <snip>;
    root /usr/share/nginx/html;
    ssl_certificate /etc/nginx/conf.d/<snip>;
    ssl_certificate_key /etc/nginx/conf.d/<snip>;

    location /<snip>/ {
        proxy_pass http://127.0.0.1:12002/;

    }
}
```

Where `self.__host` is a publicly addressable hostname.

robnagler: listen and server_name should be the same. The local IP is fine for routing.

I don't know why this would have anything to do with "in_dev_mode". In dev of rsconf you would specify the host to be something like rsiviz.v.radia.run and modify our dev aliases to support that. In prod, you'd just specify a host. You can check if the host maps to a public ip in DNS. This is a good thing in general to do (we don't currently [check that names appear in DNS](https://github.com/radiasoft/rsconf/issues/357)). 

`self.__host` will not be visible. I'm not sure why it needs to be so private. It can be part of `z`.

e-carlin: > I don't know why this would have anything to do with "in_dev_mode"

I was quite unsure how to write this code (still am). So I pieced together a frankenstein from [bivio_named](https://github.com/radiasoft/rsconf/blob/fadfa8e88d3af97a888ae0cedf992700522dea90/rsconf/component/bivio_named.py#L37) and [postfix](https://github.com/radiasoft/rsconf/blob/fadfa8e88d3af97a888ae0cedf992700522dea90/rsconf/component/postfix.py#L115).

Here is an update 141f002. Lmk if it doesn't make any sense. My thinking is on prod I'll update named to have a new alias pointing to the desired host.

e-carlin:
@robnagler ready for another review

rsconf/component/rsiviz.py
```diff
@@ -1,37 +1,46 @@
...
+:copyright: Copyright (c) 2019-2023 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
+from pykern import pkconfig
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdp
 from rsconf import component
 from rsconf import db
-from rsconf import systemd
+import socket
```
robnagler, line 12: unnecessary import

rsconf/component/rsiviz.py
```diff
@@ -1,37 +1,46 @@
...
 
 
 class T(component.T):
     def internal_build_compile(self):
         self.buildt.require_component("nginx")
         jc, z = self.j2_ctx_init()
-        self.__host = self.hdb.rsconf_db.host
         if "index_uri_secret" not in z:
             z.index_uri_secret = db.random_string()
```
robnagler, line 20: This probably wants to be saved as a secret in the db so it isn't generated each time. Otherwise, it will be hard to sync the communication sirepo. Maybe it should just be required?

robnagler (changes requested):

rsconf/component/rsiviz.py
robnagler: better

rsconf/component/rsiviz.py
e-carlin: f66ec7c

robnagler (approved):
I'll let you merge.

PR #412, opened 2023-11-14
Fix #389 off by one in sirepo/tornado_nginx.conf.jinja

### LINKED ISSUE #389: num_api_servers, number of service, and nginx routing don't add up
On alpha we have num_api_servers set to 10.

That leads to 10 services on the machine
```
sirepo# systemctl list-units --type=service --state=running | grep 'sirepo@'
sirepo@7000.service           loaded active running sirepo@{7000..7009}
sirepo@7001.service           loaded active running sirepo@{7000..7009}
sirepo@7002.service           loaded active running sirepo@{7000..7009}
sirepo@7003.service           loaded active running sirepo@{7000..7009}
sirepo@7004.service           loaded active running sirepo@{7000..7009}
sirepo@7005.service           loaded active running sirepo@{7000..7009}
sirepo@7006.service           loaded active running sirepo@{7000..7009}
sirepo@7007.service           loaded active running sirepo@{7000..7009}
sirepo@7008.service           loaded active running sirepo@{7000..7009}
sirepo@7009.service           loaded active running sirepo@{7000..7009}
```

But, nginx routing is only 0-8
```
~# grep 'server ' /etc/nginx/conf.d/<snip>conf
server {
    server 127.0.0.1:7000;
    server 127.0.0.1:7001;
    server 127.0.0.1:7002;
    server 127.0.0.1:7003;
    server 127.0.0.1:7004;
    server 127.0.0.1:7005;
    server 127.0.0.1:7006;
    server 127.0.0.1:7007;
    server 127.0.0.1:7008;
server {
```

robnagler:


Benbenbenin0 (approved):
If you have time to explain I'm curious how this bug was found, and why it wasn't a problem sooner.

e-carlin:
> If you have time to explain I'm curious how this bug was found, and why it wasn't a problem sooner.

I can't exactly remember how I found it. But, I think I was diagnosing some unrelated issue on alpha and I noticed there were only 9 routes in the nginx conf even though I knew there should be 10.

Bugs like this are hard to find. Human brains aren't great at remembering things like "is range inclusive or exclusive" (at least mine isn't). Also, it is hard to see a number in one file and then connect it to a far away file when reviewing code. In a case like this where the exact number isn't so important it is easy to just look at the code see "some" and call it good. The system was working fine so there isn't anything that would tip us off.

Here is an explanation [in a meme](https://twitter.com/secretGeek/status/7269997868?ref_src=twsrc%5Etfw%7Ctwcamp%5Etweetembed%7Ctwterm%5E7269997868%7Ctwgr%5Ee81f4fe94fb6eee3dff1ce0e49174d6ce8bf5d63%7Ctwcon%5Es1_&ref_url=https%3A%2F%2Fmartinfowler.com%2Fbliki%2FTwoHardThings.html).

robnagler:
> The system was working fine so there isn't anything that would tip us off.

The off-by-one was the ability to answer 10% fewer concurrent requests, and even then, each tornado server probably can handle more than one request. That's why we wouldn't seen it as a problem.

System sizing is a notoriously difficult problem. From this particular perspective, without something like USPAS going on, we have more than enough services to handle our concurrent request load.

Benbenbenin0:
@robnagler @e-carlin thanks for the explanation, Evan thanks for the chuckle.

PR #414, opened 2023-11-25
Fix #413 Added base_os.pam_duo

### LINKED ISSUE #413: support duo

robnagler:
@e-carlin This is top priority, thanks.

**Be sure to fixup squash merge text**

- Add component.install_resource2 to reduce repetition of resource installs
- Add component.install_rpm_key to add GPG keys to rpm db (awkward interface)
- Fix db.resource_path (was returning str)
- base_os always installs /etc/pam.d/sshd so can revert duo


rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -291,6 +291,20 @@ rsconf_install_perl_rpm() {
     rsconf_service_file_changed "$rpm_file"
 }
 
+rsconf_install_rpm_key() {
+    # installs a custom rpm_key if not already installed
+    declare rpm_key=$1
+    if rpm -q "$rpm_key" &> /dev/null; then
+        return
+    fi
+    install_download "$rpm_key" > "$tmp"
```
e-carlin, line 300: $tmp doesn't look like it is defined?

rsconf/component/__init__.py
```diff
@@ -213,6 +213,59 @@ def install_resource(self, name, j2_ctx, host_path=None):
...
+            self.install_directory(host_d)
+        if access is not None:
+            if isinstance(access, str):
+                access = PKDict(mode=access)
+            self.install_access(**access)
+        return self.install_resource(
+            f"{self.module_name}/{basename}", self.j2_ctx, host_f
+        )
+
+    def install_rpm_key(self, rpm_key, channel=None):
```
e-carlin, line 260: channel looks unused?

e-carlin (changes requested):

rsconf/package_data/rsconf/rsconf.sh
robnagler: I had forgot to push.

PR #422, opened 2023-12-05
Fix #421 github_bkp requires org arg

### LINKED ISSUE #421: github_bkp requires org param

robnagler:


Benbenbenin0 (commented):
Having trouble figuring out why github_bkp needs an org arg. Is this a dir naming/security thing?

Benbenbenin0:
Got it, thanks for the explanation

PR #429, opened 2024-01-10
Fix #428 add script.timers

### LINKED ISSUE #428: add script.timers

robnagler:
- script.bash is deprecated in favor of script.rsconf_bash
- rsconf nginx config: turn off keepalive


e-carlin (approved):
Looks good. Somewhere (ex merge message) can you please say why we're setting keepalive to 0.

PR #441, opened 2024-02-05
Fix #440: various fixes for vm_devbox systemd

### LINKED ISSUE #440: vm_devbox: changes to service file so systemd can manage the service
- [x] remove PrivateTmp (this was causing many problems)
- [x] Type=forking (the vm process forks and stays running after vagrant up exits)
- [x] vagrant reload after setting forwarded_port (need to reload for change to take)
- [x] execStop (needs to be an abspath and needs to be in /srv/vm_devbox_user. Just use a stop script and call it)
- [x] TimeoutStartSec=10min. With Type=forking the timeout is at play

e-carlin:
- remove PrivateTmp this was causing many problems
- Type=forking the vm process forks and stays running after vagrant up exits
- vagrant reload after setting forwarded_port (need to reload for change to take)
- execStop needs to be an abspath and needs to be in /srv/vm_devbox_user. Just use a stop script and call it
- TimeoutStartSec=10min. With Type=forking the timeout is at play

### DIFF
--- a/rsconf/component/vm_devbox.py
+++ b/rsconf/component/vm_devbox.py
@@ -41,8 +41,9 @@ class T(component.T):
         z.ssh_guest_host_key_f = "/etc/ssh/host_key"
         z.ssh_guest_identity_pub_f = "/etc/ssh/identity.pub"
         z.start_f = z.run_d.join("start")
+        z.stop_f = z.run_d.join("stop")
         z.vm_hostname = f"{self._user}.{jc[self.module_name].vm_parent_domain}"
-        systemd.unit_prepare(self, self.j2_ctx, watch_files=(z.start_f,))
+        systemd.unit_prepare(self, self.j2_ctx, watch_files=(z.start_f, z.stop_f))
         self._network(jc, z)
         self._ssh(jc, z)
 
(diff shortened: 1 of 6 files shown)


rsconf/package_data/vm_devbox/stop.sh.jinja
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -eou pipefail
+
+vm_devbox_main() {
+    cd {{ this.run_d }}
```
robnagler, line 5: `cd '{{ this.run_d }}'`

rsconf/component/vm_devbox.py
```diff
@@ -56,6 +57,7 @@ def internal_build_write(self):
         self.install_directory(z.run_d)
         self.install_access(mode="500", owner=z.run_u)
         self.install_resource("vm_devbox/start.sh", host_path=z.start_f)
+        self.install_resource("vm_devbox/start.sh", host_path=z.stop_f)
```
robnagler, line 60: start.sh > stop.sh

robnagler (commented):
A bug and a format.

PR #447, opened 2024-02-06
Fix #440: configuration to make ssh more secure

### LINKED ISSUE #440: vm_devbox: changes to service file so systemd can manage the service

e-carlin:


robnagler (commented):
lgtm

PR #462, opened 2024-03-22
Fix #452: create mlflow component

### LINKED ISSUE #452: create mlflow component

e-carlin:
MLFlow Tracking Server as a service with basic-auth. It supports experiment parameter tracking but does not yet support artifact tracking.


rsconf/component/mlflow.py
```diff
@@ -0,0 +1,68 @@
...
+        z.run_d = systemd.unit_run_d(jc, self.name)
+        z.db_d = z.run_d.join(_DB_SUBDIR)
+        z.auth_conf_f = z.run_d.join(_BASIC_AUTH_CONF_F)
+        z.auth_db_f = z.run_d.join(_BASIC_AUTH_DB_F)
+        systemd.docker_unit_prepare(
+            self,
+            jc,
+            docker_exec=f"mlflow server --host {_MLFLOW_IP} --port {z.service_port} --backend-store-uri 'sqlite:///{z.db_d}/backend-store.db' --no-serve-artifacts --default-artifact-root '{z.db_d}' --app-name basic-auth",
+        )
+        for s in _SECRETSz:
```
robnagler: z is typo

rsconf/component/mlflow.py
```diff
@@ -0,0 +1,68 @@
...
+"""
+
+from pykern import pkconfig
+from pykern.pkcollections import PKDict
+from pykern.pkdebug import pkdp
+from rsconf import component
+
+
+_DB_SUBDIR = "db"
+_MLFLOW_IP = "127.0.0.1"
```
robnagler, line 14: This should come from some place as LOCAL_HOST_IP. Maybe even pykern, because there's already something like this. The point isn't that it is changing, but rather that we know all the places we hardwire IPs to local host

tests/pkcli/build_data/1.out/srv/host/v9.radia.run/000.sh
```diff
@@ -26,6 +26,7 @@ rsconf_require dovecot
 rsconf_require github_bkp
 rsconf_require jupyterhub
 rsconf_require jupyterhub_proxy
+rsconf_require mlflow
```
robnagler, line 29: This file isn't necessary to have in the saved out.

tests/pkcli/build_data/1.out/srv/host/v9.radia.run/srv/mlflow/env
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
robnagler, line 1: not sure this file is necessary unless you are testing the env var.

robnagler (approved):

tests/pkcli/build_data/1.out/srv/host/v9.radia.run/srv/mlflow/env
e-carlin: I think it is worth testing in this case. If the env var is missing the server will start fine just be broken. It will use the default user/pass which seems like a large enough security risk it is worth trying to catch.

tests/pkcli/build_data/1.out/srv/host/v9.radia.run/000.sh
e-carlin: d599c0c

rsconf/component/mlflow.py
e-carlin: e52c947

PR #475, opened 2024-04-01
Fix #474: allow vagrant localhost ssh in vm_devbox

### LINKED ISSUE #474: vm_devbox allow ssh connections from localhost
For https://github.com/radiasoft/sirepo/issues/6946

e-carlin:
Allow ssh connections from vagrant over localhost in vm_devbox. This allows services like sirepo.job_driver.sbatch to connect over localhost back in to the host via ssh to start a job_agent.


e-carlin:
@robnagler you mentioned using unique passwords for vagrant instead of the default. Is that still something you think is necessary? I'll need to test that nothing breaks but I assume we can just set the password for the vagrant user when provisioning the vm.

robnagler:
We should delete the password for the vagrant user and root. The user can then set the password when they need it. We should add a quick check on setup of Slurm that tells the user what to do (what env var to set, for example).

rsconf/component/vm_devbox.py
```diff
@@ -9,6 +9,7 @@
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdp
 from rsconf import component
+from rsconf import db
```
robnagler, line 12: We should switch to qualified imports `db` is too short.

rsconf/package_data/vm_devbox/start.sh.jinja
```diff
@@ -79,6 +79,11 @@ PasswordAuthentication no
 PermitRootLogin no
 Protocol 2
 X11Forwarding yes
+
+# SSH acces from localhost (ex for sirepo.job_driver.sbatch)
```
robnagler, line 83: Typo: access

robnagler (approved):

rsconf/package_data/vm_devbox/start.sh.jinja
e-carlin: https://github.com/radiasoft/rsconf/pull/475/commits/2eb3a2218adbc6b59588554bf0d3f8a996d68025

rsconf/component/vm_devbox.py
e-carlin: https://github.com/radiasoft/rsconf/pull/475/commits/2eb3a2218adbc6b59588554bf0d3f8a996d68025

e-carlin:
please review again with the addition of `passwd -d`

rsconf/package_data/vm_devbox/start.sh.jinja
```diff
@@ -79,12 +79,20 @@ PasswordAuthentication no
...
+ListenAddress {{ this.local_ip }}:22
+Match User {{ this.run_u }} Address {{ this.local_ip }}
+    PasswordAuthentication yes
 EOF_INSTALL
 
 echo '{{ this.ssh_identity_pub_key }}' >> '/home/{{ this.run_u }}/.ssh/authorized_keys'
 
 install --mode=400 --owner=root --group=root /dev/stdin "{{ this.ssh_guest_host_key_f }}"<<EOF_INSTALL
 {{ this.ssh_host_key }}
+
```
robnagler, line 93: Missing `EOF_INSTALL`

rsconf/package_data/vm_devbox/start.sh.jinja
```diff
@@ -79,12 +79,20 @@ PasswordAuthentication no
...
+passwd -d {{ this.root_u }}
+passwd -d {{ this.run_u }}
 EOF_INSTALL
```
robnagler, line 96: Remove

robnagler (changes requested):

rsconf/package_data/vm_devbox/start.sh.jinja
e-carlin: https://github.com/radiasoft/rsconf/pull/475/commits/f6d73932d2eed3ee4acba92e60656c604e42060a

robnagler (approved):

PR #489, opened 2024-04-29
Fix #488 remove smtp_tls_eecdh_grade = ultra

### LINKED ISSUE #488: postconf: warning: /etc/postfix/main.cf: unused parameter: smtp_tls_eecdh_grade=ultra

robnagler:


schellj:
This looks correct given the info here and the config disabling TLS < 1.2: https://www.postfix.org/FORWARD_SECRECY_README.html
> With TLS 1.2 the server needs to leave its setting of "[smtpd_tls_eecdh_grade](https://www.postfix.org/postconf.5.html#smtpd_tls_eecdh_grade)" at the default value of "auto" (earlier choices of an explicit single curve grade are deprecated). With TLS 1.3, the "[smtpd_tls_eecdh_grade](https://www.postfix.org/postconf.5.html#smtpd_tls_eecdh_grade)" parameter is not used, and curve selection is unconditionally negotiated.

@robnagler Since you assigned this to me, were you looking for me to do something further?

robnagler:
@schellj just looking for your review/approval since you had worked on this code recently.

schellj (approved):

PR #491, opened 2024-04-30
Fix #490 chown xargs --no-run-if-empty

### LINKED ISSUE #490: spamd xargs chown: missing operand after ‘root:root’

robnagler:


e-carlin:
lgtm

PR #503, opened 2024-05-17
Fix #502 added pkcli.letsencrypt

### LINKED ISSUE #60: add TLS cert expiry & revocation check
[Ericsson O2 was down](https://www.ericsson.com/en/press-releases/2018/12/update-on-software-issue-impacting-certain-customers) due to a bad TLS cert.  [Hacker News mad e many comments](https://news.ycombinator.com/item?id=18622169) but no one said "add a check to ansible or chef or puppet. There is a has_expired check in ansible, but it's not what we want, and they don't check revocation (openssl verify does this). We want an assertion a week before in rsconf.

We can have backup checks for our pingers, but that's not good enough, because not all TLS certs are configured that way.

### LINKED ISSUE #501: pkcli.tls.gen-csr with existing key

### LINKED ISSUE #502: pkcli.letsencrypt
Workflow is three stages:
- [register: done once](https://github.com/radiasoft/rsconf/files/15283700/register-py.txt)
- [generate challenge](https://github.com/radiasoft/rsconf/files/15283708/challenge-py.txt)
- [validate and get fullchain crt](https://github.com/radiasoft/rsconf/files/15283710/certify-py.txt)

The renewal process just requires the last two steps with a new CSR each time. NamedConf supports arbitrary txt records so we'll just write the same text record each time.

Renewal requires #501. We don't want to regenerate the key every time although we could.

Registration can actually occur each time, but it takes time, and it clutters things. Better to have a fixed set account for this.

robnagler:
Fix #501 add gen_csr
Fix #60 checks expiry (not revocation, which is less important)
- Add tls.read_crt_as_dict, check_expired, etc.
- Added db.global_path to manage paths better


e-carlin (approved):
Looks great. Thanks for the docs and clear code.

Once this is all stitched into our system I think a demo with the team would be good. I think it would be good for everyone to learn about how tls / acme works and for Paul and I to learn about how this is built into our systems so we can debug it.

PR #507, opened 2024-05-28
Fix #504 added tls.db_yaml; fixed _find_tls_crt

### LINKED ISSUE #504: Generate yaml from certs for tls lookup
tls.check_expiring could do this so the map is always up to date.

robnagler:


e-carlin (commented):
lgtm

PR #514, opened 2024-07-30
Fix #513: still need to write empty files

### LINKED ISSUE #513: /etc/postfix/sender_access not created
Running `radia_run vagrant-rsconf-dev` I get this error:
```
curl -L -s -S -n file:///home/vagrant/src/radiasoft/rsconf/run/srv/host/v3.radia.run//etc/postfix/sender_access
curl: (37) Couldn't open file /home/vagrant/src/radiasoft/rsconf/run/srv/host/v3.radia.run//etc/postfix/sender_access
tail: cannot open ‘/root/radia-run-install.log’ for reading: No such file or directory
Unexpected error; Install failed.
```

e-carlin:


e-carlin:
Need to fix tests and probably worth adding a new one to cover this case. I'll reopen when done.

PR #535, opened 2024-12-18
Fix #534: add kickstart http config to rsconf

### LINKED ISSUE #534: add http config for kickstart
No longer serving kickstart files over nfs. Now using http. So, add to http config

```
    location /kickstart/ {
        alias /srv/kickstart/;
    }
```

e-carlin:


rsconf/component/rsconf.py
```diff
@@ -29,6 +29,10 @@ def internal_build_compile(self):
             auth_f=nginx.CONF_D.join(PASSWD_SECRET_F),
             srv_d=jc.rsconf_db.srv_d,
             host_subdir=jc.rsconf_db.srv_host_d.basename,
+            kickstart_hosts=[
```
robnagler, line 32: I don't think this is a good idea. It should just be configuration. This assumes a lot about passwd_secret_f, which it shouldn't Nothing else assumes that except for setup_dev. It also requires the host to be added to rsconf before it has a base operating system. Rather, use a CIDR for the subnet with a list of special cases.

tests/pkcli/build1_data/1.out/srv/host/v3.radia.run/etc/nginx/conf.d/v3.radia.run.conf
```diff
@@ -0,0 +1,43 @@
...
+        auth_basic_user_file /etc/nginx/conf.d/rsconf_auth;
+        # DEBUG: return 200 $remote_user;
+        if ($remote_user != $1) {
+            return 403;
+        }
+        root /home/vagrant/src/radiasoft/rsconf/tests/pkcli/build1_work/1/srv;
+    }
+
+    location /kickstart/ {
+        allow v9.radia.run;
```
robnagler, line 35: This list will get really long with the number of hosts we have. Much better to use a CIDR.

rsconf/package_data/rsconf/nginx.conf.jinja
```diff
@@ -30,4 +30,12 @@ server {
         }
         root {{ rsconf.srv_d }};
     }
+
+    location /kickstart/ {
```
robnagler, line 34: We don't normally include the trailing slash.

robnagler (changes requested):

rsconf/component/rsconf.py
e-carlin: Use trusted nets or just add config? https://github.com/radiasoft/rsconf/pull/535/commits/c2c36928e70baab657deeb5f723116cd7bcb88a2

rsconf/package_data/rsconf/nginx.conf.jinja
robnagler: The ones above are regular expressions and we have the trick in dev of having the credentials in /srv so the names would collide with the host dirs.

rsconf/component/rsconf.py
robnagler: Add config. Let the user decide. This is security so using a default would be too much.

rsconf/package_data/rsconf/nginx.conf.jinja
e-carlin: 40b6ce1

rsconf/component/rsconf.py
e-carlin: 40b6ce1

e-carlin:
@robnagler ready for another review

robnagler (approved):

PR #537, opened 2025-01-28
Fix #536: kickstart needs alias for files to be found

### LINKED ISSUE #536: rsconf nginx kickstart location block needs alias

e-carlin:


rsconf/component/rsconf.py
```diff
@@ -12,8 +12,9 @@
 from rsconf import component
 from urllib.parse import urlparse
 
-_PASSWD_SECRET_JSON_F = "rsconf_auth.json"
 PASSWD_SECRET_F = "rsconf_auth"
+_KICKSTART_ROOT_D = "/srv/kickstart"
```
robnagler, line 16: /srv shouldn't be hardwired. Use host_run_d.

robnagler (changes requested):

rsconf/component/rsconf.py
e-carlin: 170748e

robnagler (approved):

PR #546, opened 2025-02-14
Fix #545: symlink /var/lib/libvirt

### LINKED ISSUE #545: symlink /srv/libvirt /var/lib/libvirt
libvirt stores images in /var/lib/libvirt.

e-carlin:
libvirt stores data (ex images) in /var/lib/libvirt.

### DIFF
--- a/rsconf/component/__init__.py
+++ b/rsconf/component/__init__.py
@@ -614,7 +614,7 @@ def _find_tls_crt(j2_ctx, domain):
     from rsconf import db
 
     d = j2_ctx.rsconf_db.tls_d
-    for crt, domains in j2_ctx.component.tls_crt.items():
+    for crt, domains in j2_ctx.component.get("tls_crt", PKDict()).items():
         if domain in domains:
             return d.join(crt), domains
     for s in (
--- a/rsconf/component/vm_devbox.py
+++ b/rsconf/component/vm_devbox.py
@@ -16,8 +16,8 @@ import rsconf.systemd
 _DEFAULT_VAGRANT_CPUS = 4
 # Allowable pattern enforced by vagrant
 _VM_HOSTNAME_RE = "[a-z0-9][a-z0-9.-]*"
-_LIB_VIRT_D = "libvirt"
-_VAR_LIBVIRT_D = "/var/lib/libvirt"
+_LIB_VIRT_SUB_D = "libvirt"
+_VAR_LIBVIRT_D = pkio.py_path("/var/lib/libvirt")
 
 
 class T(rsconf.component.T):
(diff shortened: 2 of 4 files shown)


rsconf/component/__init__.py
```diff
@@ -614,7 +614,7 @@ def _find_tls_crt(j2_ctx, domain):
     from rsconf import db
 
     d = j2_ctx.rsconf_db.tls_d
-    for crt, domains in j2_ctx.component.tls_crt.items():
+    for crt, domains in j2_ctx.component.get("tls_crt", PKDict()).items():
```
robnagler, line 617: Why is this change necessary?

rsconf/component/vm_devbox.py
```diff
@@ -16,6 +16,8 @@
 _DEFAULT_VAGRANT_CPUS = 4
 # Allowable pattern enforced by vagrant
 _VM_HOSTNAME_RE = "[a-z0-9][a-z0-9.-]*"
+_LIB_VIRT_D = "libvirt"
```
robnagler, line 19: This is a SUBDIR

rsconf/package_data/vm_devbox/main.sh.jinja
```diff
@@ -5,6 +5,7 @@ vm_devbox_main() {
         return
     fi
     yum-config-manager --set-enabled crb
+    ln -s '{{ this.libvirt_d }}' '{{ this.var_libvirt_d }}'
```
robnagler, line 8: Do we want this as an install_symlink?

robnagler (approved):
Not sure about a couple of things.

rsconf/component/vm_devbox.py
e-carlin: 9fa7476

rsconf/package_data/vm_devbox/main.sh.jinja
e-carlin: Nice. Didn't know about that one... 9fa7476

rsconf/component/__init__.py
robnagler: I agree that tls_crt should not be required and tls_crt_create is now defaulted as well. 

ac00b67 is a "proper" fix. I don't like implicitly defining defaults in some low level module. I like them initialized in in compile or in this case. I think defaults like this should be where all the other defaults are initialized. component doesn't get access to Host and it would break the abstraction to have db import component so I POSITed the defaults in db.

LMK

e-carlin: Ya much better fix. Thanks.

PR #684, opened 2026-07-22
Fix #683 allow for version 10 of AlmaLinux

schellj:
Nothing currently requires version 9 of AlmaLinux specifically, so the `is_almalinux9` flag can be removed in favor of `is_centos7` and the install version check can be skipped. This will allow for AlmaLinux 10 hosts.

### DIFF
--- a/rsconf/db.py
+++ b/rsconf/db.py
@@ -60,13 +60,12 @@ class _Host(PKDict):
             in dbt.base.pkunchecked_nested_get("rsconf_db.centos7_hosts", ())
         )
         # For now this works
-        self.rsconf_db.is_almalinux9 = not self.rsconf_db.is_centos7
         if self.rsconf_db.is_centos7:
             self.rsconf_db.os_release_id = "centos"
             self.rsconf_db.os_release_version_id = "7"
         else:
             self.rsconf_db.os_release_id = "almalinux"
-            self.rsconf_db.os_release_version_id = "9"
+            self.rsconf_db.os_release_version_id = ""
         pkjson.dump_pretty(self, filename=self.rsconf_db.tmp_d.join("db.json"))
         self.dbt = dbt
 
--- a/rsconf/db.py
+++ b/rsconf/db.py
@@ -12,6 +12,7 @@ from pykern import pkresource
 from pykern.pkdebug import pkdc, pkdlog, pkdp, pkdpretty
 import collections
 import copy
+import os
 import random
 import re
 import string
@@ -65,7 +66,10 @@ class _Host(PKDict):
             self.rsconf_db.os_release_version_id = "7"
         else:
             self.rsconf_db.os_release_id = "almalinux"
-            self.rsconf_db.os_release_version_id = ""
+            # TODO(schellj) update variable name to install_version_almalinux
+            self.rsconf_db.os_release_version_id = os.environ.get(
+                "install_version_centos", "9"
+            )
         pkjson.dump_pretty(self, filename=self.rsconf_db.tmp_d.join("db.json"))
         self.dbt = dbt
 
--- a/rsconf/db.py
+++ b/rsconf/db.py
@@ -66,9 +66,8 @@ class _Host(PKDict):
             self.rsconf_db.os_release_version_id = "7"
         else:
             self.rsconf_db.os_release_id = "almalinux"
-            # TODO(schellj) update variable name to install_version_rhel
             self.rsconf_db.os_release_version_id = os.environ.get(
-                "install_version_centos", "9"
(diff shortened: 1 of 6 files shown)


rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -566,7 +566,7 @@ rsconf_setup_vars() {
     if [[ $install_os_release_id != $id ]]; then
         install_err "invalid os_release_id=$install_os_release_id expecting=$id"
     fi
-    if  [[ $install_os_release_version_id != $version ]]; then
+    if  [[ $version && $install_os_release_version_id != $version ]]; then
         install_err "invalid os_release_version_id=$install_os_release_version_id expecting=$version"
     fi
 }
```
robnagler, line 572: The code above is `version=$3` so it has to be set. If it isn't, that's a serious bug. If it is blank it will not equal `$install_os_release_version_id` so the `$version &&` is unnecessary.

rsconf/db.py
```diff
@@ -60,13 +60,12 @@ def __init__(self, dbt, channel, host):
...
         )
         # For now this works
-        self.rsconf_db.is_almalinux9 = not self.rsconf_db.is_centos7
         if self.rsconf_db.is_centos7:
             self.rsconf_db.os_release_id = "centos"
             self.rsconf_db.os_release_version_id = "7"
         else:
             self.rsconf_db.os_release_id = "almalinux"
-            self.rsconf_db.os_release_version_id = "9"
+            self.rsconf_db.os_release_version_id = ""
```
robnagler: This should always be set.

rsconf/systemd.py
```diff
@@ -388,16 +388,16 @@ def _on_calendar(value, jc, now=None):
...
 
     def _tz_adjustment():
-        if jc.rsconf_db.is_almalinux9:
-            return 0
-        return (
-            -int(
-                pytz.timezone(jc.systemd.timezone)
-                .utcoffset(now or datetime.datetime.utcnow())
-                .total_seconds()
+        if jc.rsconf_db.is_centos7:
```
robnagler: The inverted test makes the code easier to read. In this version, you have to skip passed old code (centos 7) to figure out what is the normal case.

tests/pkcli/build1_data/1.out/srv/host/v5.radia.run/000.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-rsconf_setup_vars 'dev' 'almalinux' '9'
+rsconf_setup_vars 'dev' 'almalinux' ''
```
robnagler: You have to set something.

robnagler (changes requested):

rsconf/db.py
schellj: How do we know what the version is? As mentioned in the issue, I was trying to avoid adding something like `almalinux10_hosts` for each new supported OS version. Also, as far as I'm aware there isn't a need to differentiate between 9 and 10.

robnagler: When you set up a host, you have to say what the host you are installing is getting. `install_version_centos` will be set so you can use that. For now, we can assume the primary os is the same as the install os's. If it isn't set (bootstrap), you would set it for the initial setup or put it in `/etc/profile.d/rs-redhat-base.sh` on the primary. I'm not sure if this is included in the bootstrap instructions.

schellj: Is this what you mean? 3ce1a736a25cef8dc6ccb687f8b7c3d4878332e0

Also, I can update the env var name to `install_version_almalinux` now if you want, or punt. I see that `install_version_centos` is used for the almalinux version at least one other place (in radiasoft/download).

rsconf/systemd.py
schellj: Resolved with 0c4b22c459bc74e14e0afde5864920748210293d.

rsconf/db.py
robnagler: Yes.

rsconf/package_data/rsconf/rsconf.sh
schellj: Resolved with 3ce1a736a25cef8dc6ccb687f8b7c3d4878332e0.

tests/pkcli/build1_data/1.out/srv/host/v5.radia.run/000.sh
schellj: Resolved with 3ce1a736a25cef8dc6ccb687f8b7c3d4878332e0.

robnagler (approved):

PR #686, opened 2026-08-05
fix #685 install modules required for iptables --mask in almalinux 10

schellj:


rsconf/package_data/base_os/main.sh.jinja
```diff
@@ -26,6 +26,15 @@ base_os_journal_persist() {
     rsconf_service_trigger_restart systemd-journald
 }
 
+base_os_kernel_modules_extra() {
+{% if rsconf_db.os_release_id == "almalinux" and rsconf_db.os_release_version_id == "10" %}
```
robnagler: This type of test should always be `>=`. This code seems in the wrong place. If you don't have iptables turned on, then you shouldn't need extra modules. I would put this in the network component. The test should probably be a feature test if possible (modinfo ip_tables) instead of a version test. Finally, when you end up with a ":" statement, may be better to modularize the code differently. You could have the conditional at the call. The embedding of the conditional in this function makes the code harder to read than if the conditional were around the function. From a modularity perspective, it's at the jinja level so the entire module is a subject to the jinja pass before it because a self-contained bash program. The modularity is therefore at the jinja level, and the goal is to have that code as readable as possible.

tests/pkcli/build1_data/1.out/srv/host/v3.radia.run/base_os.sh
```diff
@@ -56,6 +56,10 @@ base_os_journal_persist() {
     rsconf_service_trigger_restart systemd-journald
 }
 
+base_os_kernel_modules_extra() {
```
robnagler: When adding features like this, it's good if you add a test case. That's one of my AI rules:

- When beginning a bug fix, first write a test case that reproduces the problem before writing the fix. Discuss the reproduction approach with the user.

robnagler (changes requested):

tests/pkcli/build1_data/1.out/srv/host/v3.radia.run/base_os.sh
schellj: Yeah, I had commented that I was going to add a test. I need to spend some time on my AI rules/config. That looks like a good one, though Claude often adds tests on it's own.

robnagler (approved):

PR #688, opened 2026-08-11
Fix #687 Fix postgresql log install when old log dir is empty

### LINKED ISSUE #687: Postgresql install fails if old log dir doesn't exist
If logrotate is being installed, then https://github.com/radiasoft/rsconf/blob/master/rsconf/package_data/logrotate/main.sh.jinja#L2 enables nullglob, meaning that if the old log dir doesn't exist, then https://github.com/radiasoft/rsconf/blob/master/rsconf/package_data/postgresql/main.sh.jinja#L32 turns into just `for f in $(ls -tr); do...`, and if there is a subdir in the current directory, then the `cat` will throw an error:
```
Source: /var/tmp/radia-run-82288-14189/20260809005355-postgresql.sh
cat: etc: Is a directory
bash stack:
  postgresql_log
  postgresql_main
  postgresql_rsconf_component
  rsconf_run
  rsconf_require
  source
  install_script_eval
  rsconf_main
  install_script_eval
  install_repo_internal
  install_main
```

schellj:
The existing logic tries to save existing log files, but is insufficient when the assumed old log dir doesn't exist or is empty and the logrotate component is also being installed (the logic for which would come before the postgresql logic).

One or both of the commits here could be employed to handle this issue. Could also just not try to save the old logs, or fix this in another way.

### DIFF
--- a/rsconf/package_data/logrotate/main.sh.jinja
+++ b/rsconf/package_data/logrotate/main.sh.jinja
@@ -1,5 +1,4 @@
 #!/bin/bash
-shopt -s nullglob
 
 logrotate_main() {
     # We use a systemd timer so can have output in the journal, and not emailed (anacron requires email)
@@ -17,7 +16,8 @@ logrotate_main() {
             's/\brotate/#rotate/'
     fi
     for f in /etc/logrotate.d/*; do
-        if grep --no-messages --quiet ' delaycompress' "$f"; then
+        # without nullglob f could be /etc/logrotate.d/* literal
+        if [[ -f $f ]] && grep --no-messages --quiet ' delaycompress' "$f"; then
             rsconf_edit_no_change_res=0 rsconf_edit "$f" '#delaycompress' \
                's/\bdelaycompress/#delaycompress/'
         fi
--- a/rsconf/package_data/logrotate/main.sh.jinja
+++ b/rsconf/package_data/logrotate/main.sh.jinja
@@ -16,8 +16,7 @@ logrotate_main() {
             's/\brotate/#rotate/'
     fi
     for f in /etc/logrotate.d/*; do
-        # without nullglob f could be /etc/logrotate.d/* literal
-        if [[ -f $f ]] && grep --no-messages --quiet ' delaycompress' "$f"; then
+        if grep --no-messages --quiet ' delaycompress' "$f"; then
             rsconf_edit_no_change_res=0 rsconf_edit "$f" '#delaycompress' \
                's/\bdelaycompress/#delaycompress/'
         fi
--- a/rsconf/package_data/postgresql/main.sh.jinja
+++ b/rsconf/package_data/postgresql/main.sh.jinja
@@ -29,8 +29,11 @@ postgresql_log() {
     if [[ ! -L $old ]]; then
         install -d -m 700 '{{ postgresql.log_d }}'
         # save the old logs, just in case
-        for f in $(ls -tr "$old"/*.log 2>/dev/null || true); do
-            cat $f >> '{{ postgresql.log_f }}'
+        # don't use glob with ls to prevent listing cwd if nullglob and old is missing/empty
+        for f in $(ls -tr "$old" 2>/dev/null || true); do
+            if [[ $f == *.log ]]; then
+                cat "$old/$f" >> '{{ postgresql.log_f }}'
+            fi
         done
         chown -R postgres:postgres '{{ postgresql.log_d }}'
         rm -rf "$old"
--- a/rsconf/package_data/postgresql/main.sh.jinja
+++ b/rsconf/package_data/postgresql/main.sh.jinja
@@ -26,6 +26,9 @@ postgresql_install() {
 
 postgresql_log() {
     local old=/var/lib/pgsql/data/pg_log
+    if [[ '{{ postgresql.log_d }}' == "$old" ]]; then
+        install_err "postgresql.log_d must not be $old, which is replaced with a symlink"
+    fi
     if [[ ! -L $old ]]; then
         install -d -m 700 '{{ postgresql.log_d }}'
         # save the old logs, just in case
--- a/rsconf/package_data/postgresql/main.sh.jinja
+++ b/rsconf/package_data/postgresql/main.sh.jinja
@@ -29,12 +29,7 @@ postgresql_log() {
     if [[ ! -L $old ]]; then
         install -d -m 700 '{{ postgresql.log_d }}'
         # save the old logs, just in case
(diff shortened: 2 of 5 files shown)


rsconf/package_data/logrotate/main.sh.jinja
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-shopt -s nullglob
```
robnagler, line 2: Please leave this.

rsconf/package_data/logrotate/main.sh.jinja
```diff
@@ -17,7 +16,8 @@ logrotate_main() {
             's/\brotate/#rotate/'
     fi
     for f in /etc/logrotate.d/*; do
-        if grep --no-messages --quiet ' delaycompress' "$f"; then
+        # without nullglob f could be /etc/logrotate.d/* literal
+        if [[ -f $f ]] && grep --no-messages --quiet ' delaycompress' "$f"; then
```
robnagler: The [[ ]] was required because the shopt was removed.

rsconf/package_data/postgresql/main.sh.jinja
```diff
@@ -29,8 +29,11 @@ postgresql_log() {
     if [[ ! -L $old ]]; then
         install -d -m 700 '{{ postgresql.log_d }}'
         # save the old logs, just in case
-        for f in $(ls -tr "$old"/*.log 2>/dev/null || true); do
```
robnagler, line 32: The `ls -ltr` is incorrect. The correct line which fixes everything is:

`cat  "$old"/*.log /dev/null >> '{{ postgresql.log_f }}'`

If there are no logs, nothing will be appended. The log will be touched, but that's not a big deal.

rsconf/package_data/postgresql/main.sh.jinja
```diff
@@ -29,8 +29,11 @@ postgresql_log() {
...
-            cat $f >> '{{ postgresql.log_f }}'
+        # don't use glob with ls to prevent listing cwd if nullglob and old is missing/empty
+        for f in $(ls -tr "$old" 2>/dev/null || true); do
+            if [[ $f == *.log ]]; then
```
robnagler: Remove this `if`

rsconf/package_data/postgresql/main.sh.jinja
```diff
@@ -29,8 +29,11 @@ postgresql_log() {
     if [[ ! -L $old ]]; then
         install -d -m 700 '{{ postgresql.log_d }}'
```
robnagler, line 33: Assumes that `{{ postgresql.log_d }} != $old`, which should be asserted. Also, `postgresql_log` will never exist, because `postgresql_install` moves `$old` to `run_d`. Ask the AI about this.

robnagler (changes requested):

rsconf/package_data/logrotate/main.sh.jinja
schellj: Will do, but we're ok with setting this here when it can affect the components that are included after logrotate?

rsconf/package_data/postgresql/main.sh.jinja
schellj: Thanks. Claude and I went back and forth on the fix and it wanted to use overly complicated solutions to the potential nullglob issue.

This does lose the sorting by modified time, but looking now and there's probably only ever one *.log file and I guess it's not a big deal if they get appended out of order.

rsconf/package_data/logrotate/main.sh.jinja
robnagler: Ideally everything should be be nullglob. Have claude review that this is ok and put the nullglob in the main, and alternative is to bracket nullglob in logrotate_main, which seems suboptimal.

rsconf/package_data/postgresql/main.sh.jinja
robnagler: `*.log` should retain the date order afaik.

robnagler: A good rule (especially with LLMs) is: bug fixes should reduce code, not add code. The point of nullglob, for example, is to remove code like `== *.log`.

schellj: > `*.log` should retain the date order afaik.

Would be lexical, but I don't think it really matters here.

rsconf/package_data/logrotate/main.sh.jinja
schellj: Claude says there are currently only two sites affected by nullglob behavior (https://github.com/radiasoft/rsconf/blob/master/rsconf/package_data/logrotate/main.sh.jinja#L18, and https://github.com/radiasoft/rsconf/blob/master/rsconf/package_data/rsconf/rsconf.sh#L591) and each work as intended with it enabled, so that's fine, but it complains that it can result in silent failures like the one in the linked issue. Also, the global nullglob setting would affect root's .bashrc (which could theoretically contain code not controlled by rsconf) because is sourced by base_users.py (https://github.com/radiasoft/rsconf/blob/687-fix-postgresql-log-install/rsconf/component/base_users.py#L70).

So, global nullglob appears to be fine with the current codebase, but there is potential for subtle bugs.

robnagler: Great. Let's just move it into the place where the `set -eou pipefail` is.

> result in silent failures like the one in the linked issue

It wasn't a silent failure. :)

schellj: > Great. Let's just move it into the place where the set -eou pipefail is.

ok, will do.

> It wasn't a silent failure. :)

I guess, but I could imagine other scenarios where non-error-causing issues result from the glob resolving to a missing command argument.

schellj: Like this issue only caused an error because there was a directory in the cwd; if that wasn't the case, it would have silently catted unrelated files into the pgsql log.

rsconf/package_data/postgresql/main.sh.jinja
schellj: Added the assertion. However, I don't think your second point is correct; `postgresql_install` moves `$old` to `run_d`, but then creates a symlink from `$old` to `run_d`, so if there was originally a `/var/lib/pgsql/data/pg_log`, then it would still be accessible through the symlink.

robnagler: Agreed.

robnagler (approved):

Issue #691, opened 2026-08-17
spamd install fails if /etc/mail doesn't already exist


schellj:
```
Source: /var/tmp/radia-run-200153-6111/20260817182219-spamd.sh
bash stack:
  rsconf_install_directory
  spamd_rsconf_component
  rsconf_run
  rsconf_require
  source
  install_script_eval
  rsconf_main
  install_script_eval
  install_repo_internal
  install_main
/etc/mail/spamassassin: parent directory (/etc/mail) does not exist
```

PR #694, opened 2026-08-17
Fix #693 sa_update_keys_d created by sa-update, may not have run yet

### LINKED ISSUE #693: spamassassin sa-update-keys dir may not exist (yet)
```
find: ‘/etc/mail/spamassassin/sa-update-keys’: No such file or directory
bash stack:
  spamd_main
  spamd_rsconf_component
  rsconf_run
  rsconf_require
  source
  install_script_eval
  rsconf_main
  install_script_eval
  install_repo_internal
  install_main
```

schellj:


robnagler (changes requested):
I think sa-update should be run manually if the keys don't exist. They keys are required for normal operation.

schellj:
In general, are we to presume that component dependencies are installed, either by a prior component, or by the component itself? If so, it appears that the root cause here and for #691 is that spamassassin hasn't been installed yet. On my host, what installs container-perl (which installs spamassassin) is `artisans.sh`, and in `000.sh`, `rsconf_require spamd` comes before `rsconf_require artisans`. Looks like the potential ordering issue results from some older components not using `internal_build_compile` and the write queue.

robnagler:
> In general, are we to presume that component dependencies are installed, either by a prior component, or by the component itself? 

Yes

> If so, it appears that the root cause here and for #691 is that spamassassin hasn't been installed yet.

Something is not right, because spamd does install bivio-perl rpm. The build_compile and build_write phases happen on the rsconf controller, not the target. The error is on the target. 

Look a bit deeper into why this is happening. bivio-perl may be a dummy in rsconf dev mode.

schellj:
I think the problem is that the bop component uses `internal_build`, while `spamd` uses `internal_build_compile`/`internal_build_write`. 

The calls to `internal_build` happen at compile time (https://github.com/radiasoft/rsconf/blob/693-sa-update-keys-may-not-exist/rsconf/component/__init__.py#L78); calls to `internal_build_write` get deferred via the write queue (https://github.com/radiasoft/rsconf/blob/693-sa-update-keys-may-not-exist/rsconf/component/__init__.py#L76) to after all of the components have been compiled.

This means that when the component for the first bop app is compiled, `install_perl_rpms` is called (https://github.com/radiasoft/rsconf/blob/693-sa-update-keys-may-not-exist/rsconf/component/bop.py#L62), which adds the call to `install_repo_eval biviosoftware/container-perl base` (https://github.com/radiasoft/rsconf/blob/693-sa-update-keys-may-not-exist/rsconf/component/bop.py#L160) to the component script and sets `bop._perl_installed = True`. This prevents the call to `install_repo_eval biviosoftware/container-perl base` from being added to the script for further components that want to `install_perl_rpms` (in their `internal_build_write`), like spamd (https://github.com/radiasoft/rsconf/blob/693-sa-update-keys-may-not-exist/rsconf/component/spamd.py#L39). 

The component ordering in `000.sh` means that the spamd script gets called before the first bop script, and before spamassassin is installed via container-perl.

schellj:
It looks like the simplest potential workaround would be to drop the `_perl_installed` gate from `install_perl_rpms`; there would then be redundant `container-perl` install attempts, but those should be idempotent.

However, there is also a similar install ordering issue in postgrey with `bivio-perl`. A similar workaround would be to remove the `_installed_rpms` gate from `install_perl_rpm` here: https://github.com/radiasoft/rsconf/blob/693-sa-update-keys-may-not-exist/rsconf/component/__init__.py#L209

Looks like there are other more ideal solutions that would be more involved.

robnagler:
The coupling of install_perl_rpms inside bop is incorrect. That was laziness. The correct solution would be to have a component: perl_rpms (or other name). This would be required by spamd, bop, and postgrey. This would fix the ordering problem and the state management. A component is only installed once, and before anything that requires it.

schellj:
ok, that's actually the recommended one of two other solutions that Claude proposed. The other being to refactor the components that still use `internal_build` to use `internal_build_compile`/`internal_build_write`.

robnagler:
The build_compile/build_write fixes would be fine but still obscure the dependency relationship. The fundamental issue is a run once flag (gate is poor usage, as chatgpt realized after a brief discussion :) should be bound to the entity that's doing the work, not to an arbitrary entity.

schellj:
ok, added the component (`bop_perl`).

One question, though; is it weird that spamd and postgrey require bop rpms? Mightn't you want use those without having any bop apps?

robnagler:
We could split the rpms into parts. That's work, and the space the rpms take isn't that much. Another option is to require the rpms explicitly in spamd & postgrey, and then this problem would go away. I was just trying to do the simplest thing, but maybe the latter is better especially since we can use AI to figure out which RPMs are required.

robnagler (changes requested):
This needs to be simplified. I wrote a comment. Tell Claude to remove the bop depednency.

robnagler (approved):

PR #696, opened 2026-08-21
Fix #695 Use unix socket to work around TCP milter processing delays

### LINKED ISSUE #695: Postfix opendkim milter processing is slow
Tested on a beta host, see below. With the opendkim milter enabled (first call to sendmail), the for a 33 MB email, the pre-queue delay was 23 seconds; without it was 1.3 seconds. The delay grows linearly with the email size.
```
~# { printf 'Subject: dkim timing test\n\n'; base64 /dev/urandom | head -c 33000000; } > /tmp/big.eml
~# /usr/sbin/sendmail -i test_dkim_on@example.com < /tmp/big.eml
~# postconf -e 'non_smtpd_milters=' && postfix reload
postfix/postfix-script: refreshing the Postfix mail system
~# /usr/sbin/sendmail -i test_dkim_off@example.com < /tmp/big.eml
~#
~# grep test_dkim /var/log/maillog
postfix/smtp[33079]: 2C2E580C4E4: to=<test_dkim_on@example.com>, relay=none, delay=23, delays=23/0.06/0.03/0, dsn=5.4.4, status=bounced (Name service error for name=example.com type=MX: Malformed or unexpected name server reply)
postfix/smtp[33079]: 831688130A0: to=<test_dkim_off@example.com>, relay=none, delay=1.3, delays=1.3/0/0.03/0, dsn=5.4.4, status=bounced (Name service error for name=example.com type=MX: Malformed or unexpected name server reply)
```
It appears that the delay is related to the fact that postfix sends message content to milters in a hardcoded 64 KB chunk size, and the current opendkim component configuration is to use TCP via localhost. See https://en.wikipedia.org/wiki/Nagle%27s_algorithm#Interaction_with_delayed_ACK. A possible workaround would be to use a unix socket instead.

schellj:
Open to other workarounds for #695. I think the main other route may have been to use TCP_NODELAY, but it appears that would have to be a enabled on the host itself as there isn't a postfix milter setting for that.

### DIFF
--- a/rsconf/component/opendkim.py
+++ b/rsconf/component/opendkim.py
@@ -12,6 +12,7 @@ from rsconf import component
 
 _CONF_D = pkio.py_path("/etc/opendkim")
 _CONF_F = _CONF_D.new(ext=".conf")
+_RUN_D = pkio.py_path("/run/opendkim")
 _SECRET_SUBDIR = "opendkim"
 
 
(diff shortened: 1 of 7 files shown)


rsconf/component/opendkim.py
```diff
@@ -20,14 +21,19 @@ def internal_build_compile(self):
         jc, z = self.j2_ctx_init()
         self.buildt.require_component("postfix")
         self.append_root_bash("rsconf_yum_install opendkim")
+        # TODO: remove port config. the milter changed to use a unix socket. Accepted so hosts
+        # that still set it don't fail to build.
```
robnagler: I don't understand the comment. Just remove port=8891

robnagler (changes requested):

robnagler (approved):

PR #698, opened 2026-08-25
Fix #697 Need unit override to set required opendkim socket permissions

### LINKED ISSUE #697: postfix permission denied on opendkim unix socket
```
postfix/cleanup[18316]: warning: connect to Milter service unix:/run/opendkim/opendkim.sock: Permission denied
```

```
~# ls -ld /run/opendkim
drwxrwxr-x 2 opendkim opendkim 80 Aug 25 15:26 /run/opendkim
~# ls -l /run/opendkim/opendkim.sock
srwxrwx--- 1 opendkim opendkim 0 Aug 25 15:26 /run/opendkim/opendkim.sock
```

schellj:


rsconf/component/opendkim.py
```diff
@@ -45,6 +45,9 @@ def internal_build_write(self):
         z = jc[self.name]
         systemd.unit_prepare(self, jc, watch_files=(_CONF_D, _CONF_F))
         self._install_conf(jc, z, self._install_keys(jc, z))
+        # the unit sets Group so opendkim starts unprivileged and cannot honor
+        # UserID's group, which is what gives the socket its group
+        systemd.install_unit_override(self, jc)
```
robnagler: This is backwards. The postfix group holds mail (PII) and opendkim should not have privileges to access postfix. This is a (big) security hole if there's a bug in opendkim (more likely than postfix, which is older and more broadly used albeit more complex). 

I like this solution: https://wiki.gentoo.org/wiki/OpenDKIM. It's more involved, but we're using AI to write this so should be easy, and I'm sure Claude will tell you this is a more secure solution. :)

schellj: Actually, Claude doesn't think that's a more secure solution:
>   So the dkimsocket indirection buys nothing on this system. Gentoo's dedicated group is defending against
  an exposure that Postfix's directory-based gating already prevents. I'd leave Group=postfix as written.

>   The one thing I'd still adopt from that page is the stricter key ownership (root:opendkim 0640 instead of
  our opendkim:opendkim 0400), so a compromised daemon can't rewrite its own signing keys. That's a
  separate change and it's coupled to the group decision, so it's worth its own issue rather than bolting
  onto this one.

robnagler: When the group of the process is `postfix` the process can read all group read postfix files. I don't know what those are, but the default umask is 022 on RH except we change it to 027. 

The opendkim keys [are already 400](https://github.com/radiasoft/rsconf/blob/f276ec4bab9969ccc0570aae27445113e89488c9/rsconf/component/opendkim.py#L79).

Curious what your session says...

robnagler (changes requested):

schellj:
After discussion, closing in favor of #701.

PR #701, opened 2026-08-29
Fix #695 opendkim unix socket milter

### LINKED ISSUE #695: Postfix opendkim milter processing is slow

schellj:
Behind feature flag so hosts opt-in to using a unix socket for the opendkim postfix milter. Socket uses `dkimsocket` group, to which `postfix` is added and `opendkim` unit is run with.

Planning to review changes again, so PR is still draft status.

### DIFF
--- a/rsconf/component/opendkim.py
+++ b/rsconf/component/opendkim.py
@@ -12,6 +12,7 @@ from rsconf import component
 
 _CONF_D = pkio.py_path("/etc/opendkim")
 _CONF_F = _CONF_D.new(ext=".conf")
+_RUN_D = pkio.py_path("/run/opendkim")
 _SECRET_SUBDIR = "opendkim"
 
 
@@ -20,14 +21,19 @@ class T(component.T):
         jc, z = self.j2_ctx_init()
         self.buildt.require_component("postfix")
         self.append_root_bash("rsconf_yum_install opendkim")
+        # TODO: remove port config. the milter changed to use a unix socket. Accepted so hosts
+        # that still set it don't fail to build.
         z.pksetdefault(port=8891, smtp_clients=[])
         z.update(
             external_ignore_list_f=_CONF_D.join("ExternalIgnoreList"),
             internal_hosts_f=_CONF_D.join("InternalHosts"),
             key_table_f=_CONF_D.join("KeyTable"),
             keys_d=_CONF_D.join("keys"),
+            # postfix connects to socket_f so must be able to write it
+            run_g="postfix",
             run_u="opendkim",
             signing_table_f=_CONF_D.join("SigningTable"),
+            socket_f=_RUN_D.join("opendkim.sock"),
         )
         self._read_keys(jc, z)
         self.buildt.get_component("postfix").setup_opendkim(self)
--- a/rsconf/component/opendkim.py
+++ b/rsconf/component/opendkim.py
@@ -14,6 +14,7 @@ _CONF_D = pkio.py_path("/etc/opendkim")
 _CONF_F = _CONF_D.new(ext=".conf")
 _RUN_D = pkio.py_path("/run/opendkim")
 _SECRET_SUBDIR = "opendkim"
+_SOCKET_G = "dkimsocket"
 
 
 class T(component.T):
@@ -21,18 +22,21 @@ class T(component.T):
         jc, z = self.j2_ctx_init()
         self.buildt.require_component("postfix")
         self.append_root_bash("rsconf_yum_install opendkim")
-        z.pksetdefault(smtp_clients=[])
+        z.pksetdefault(
+            port=8891,
+            smtp_clients=[],
+            socket_gid=3000,
+            want_unix_socket=False,
+        )
         z.update(
             external_ignore_list_f=_CONF_D.join("ExternalIgnoreList"),
             internal_hosts_f=_CONF_D.join("InternalHosts"),
             key_table_f=_CONF_D.join("KeyTable"),
             keys_d=_CONF_D.join("keys"),
-            # postfix connects to socket_f so must be able to write it
-            run_g="postfix",
             run_u="opendkim",
             signing_table_f=_CONF_D.join("SigningTable"),
-            socket_f=_RUN_D.join("opendkim.sock"),
         )
+        self._setup_socket(z)
         self._read_keys(jc, z)
         self.buildt.get_component("postfix").setup_opendkim(self)
         self._trusted_hosts(jc, z)
@@ -45,6 +49,11 @@ class T(component.T):
         z = jc[self.name]
         systemd.unit_prepare(self, jc, watch_files=(_CONF_D, _CONF_F))
         self._install_conf(jc, z, self._install_keys(jc, z))
+        if z.want_unix_socket:
+            self.append_root_bash_with_main(jc)
+            # the unit sets Group so opendkim starts unprivileged and cannot
+            # honor UserID's group, which is what gives the socket its group
+            systemd.install_unit_override(self, jc)
         systemd.unit_enable(self, jc)
 
     def _install_conf(self, jc, z, key_list):
@@ -123,6 +132,27 @@ class T(component.T):
                 raise AssertionError(f"missing keys for domain={d}")
             for k in v.rows:
                 k.subdomain = opendkim.public_key_info(k.txt_f)
+
+    def _setup_socket(self, z):
+        if not z.want_unix_socket:
+            z.update(
+                milter=f"inet:localhost:{z.port}",
+                run_g=z.run_u,
+                socket=f"inet:{z.port}@localhost",
+                socket_f=None,
+            )
+            return
+        f = _RUN_D.join("opendkim.sock")
+        # f is created with run_g so socket_client_u can connect to it. A group
+        # of its own keeps socket_client_u out of the group that owns keys_d,
+        # and run_u out of the group that owns the postfix queue.
+        z.update(
+            milter=f"unix:{f}",
+            run_g=_SOCKET_G,
+            socket=f"local:{f}",
+            socket_client_u="postfix",
+            socket_f=f,
+        )
 
     def _trusted_hosts(self, jc, z):
         from rsconf import db
--- a/rsconf/component/opendkim.py
+++ b/rsconf/component/opendkim.py
@@ -21,9 +21,7 @@ class T(component.T):
         jc, z = self.j2_ctx_init()
         self.buildt.require_component("postfix")
         self.append_root_bash("rsconf_yum_install opendkim")
-        # TODO: remove port config. the milter changed to use a unix socket. Accepted so hosts
-        # that still set it don't fail to build.
-        z.pksetdefault(port=8891, smtp_clients=[])
+        z.pksetdefault(smtp_clients=[])
         z.update(
             external_ignore_list_f=_CONF_D.join("ExternalIgnoreList"),
             internal_hosts_f=_CONF_D.join("InternalHosts"),
--- a/rsconf/package_data/opendkim/main.sh.jinja
+++ b/rsconf/package_data/opendkim/main.sh.jinja
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+opendkim_main() {
+    rsconf_group '{{ opendkim.run_g }}' '{{ opendkim.socket_gid }}'
+    if id -nG '{{ opendkim.socket_client_u }}' \
+        | grep -F -q -w -- '{{ opendkim.run_g }}'; then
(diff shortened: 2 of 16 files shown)


rsconf/component/opendkim.py
```diff
@@ -12,15 +12,22 @@
 
 _CONF_D = pkio.py_path("/etc/opendkim")
 _CONF_F = _CONF_D.new(ext=".conf")
+_RUN_D = pkio.py_path("/run/opendkim")
 _SECRET_SUBDIR = "opendkim"
+_SOCKET_G = "dkimsocket"
```
robnagler: `_SOCKET_GROUP`. The `<name>_D` is used a lot so deserved a shorter prefix. `_G` could mean anything.

rsconf/component/opendkim.py
```diff
@@ -29,6 +36,7 @@ def internal_build_compile(self):
             run_u="opendkim",
             signing_table_f=_CONF_D.join("SigningTable"),
         )
+        self._setup_socket(z)
```
robnagler: In build_write it's an explicit conditional. In this case, not. Probably better to pick one or the other.

rsconf/component/opendkim.py
```diff
@@ -120,6 +133,27 @@ def _find(key_d):
             for k in v.rows:
                 k.subdomain = opendkim.public_key_info(k.txt_f)
 
+    def _setup_socket(self, z):
+        if not z.want_unix_socket:
```
robnagler: The z.update is duplicated.  The internal values (socket_f, socket, milter) probably should be underscore values.

rsconf/package_data/opendkim/main.sh.jinja
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+opendkim_main() {
+    rsconf_group '{{ opendkim.run_g }}' '{{ opendkim.socket_gid }}'
```
robnagler: The run_g could be run_gid or run_group. It's shouldn't just be "g"

rsconf/component/opendkim.py
```diff
@@ -12,15 +12,22 @@
...
 
 
 class T(component.T):
     def internal_build_compile(self):
         jc, z = self.j2_ctx_init()
         self.buildt.require_component("postfix")
         self.append_root_bash("rsconf_yum_install opendkim")
-        z.pksetdefault(port=8891, smtp_clients=[])
+        z.pksetdefault(
+            port=8891,
```
robnagler: seems like this should be in setup_socket. The port is unnecessary here. It should be defaulted in setup_socket only if want_unix_socket is false. The same with socket_gid.

rsconf/component/opendkim.py
```diff
@@ -41,6 +49,11 @@ def internal_build_write(self):
         z = jc[self.name]
         systemd.unit_prepare(self, jc, watch_files=(_CONF_D, _CONF_F))
         self._install_conf(jc, z, self._install_keys(jc, z))
+        if z.want_unix_socket:
+            self.append_root_bash_with_main(jc)
+            # the unit sets Group so opendkim starts unprivileged and cannot
```
robnagler: This comment doesn't make sense. You have to set group and user since there's no way to set that in the opendkim config. This is somewhat important, because it still has opendkim as its group for file ownership (if say, we wanted files owned by root and read by opendkim group). That works automatically in systemd afaik. Otherwise, postfix wouldn't also get the dkimsocket group.

rsconf/component/opendkim.py
robnagler: This shortened name is poor naming. All names should contain opendkim. "dkim" means something else (protocol). It's opendkim's socket.

Now that I see the implementation. It now seems that this should be the normal opendkim group, not a new group. It's much more complicated this way (SOCKET_G is never created -- see below). Giving postfix opendkim group privs is probably just fine from a security point of view. I noticed ArchLinux does it the other way which is where Claude probably got that solution. That's definitely not the way to do it as opendkim is a service to postfix (and postfix only).

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -139,12 +139,12 @@ rsconf_group() {
         fi
         return
     fi
-    declare flags=()
```
robnagler: This change is wrong and not the way we do it elsewhere. The only bug is the < 1000 which should be `(())`. `system_flag` is bizarre. Not sure what `system` is, but it's simply a flag and should be an array so you can always quote the arguments correctly.

rsconf/package_data/rsconf/rsconf.sh
```diff
@@ -640,25 +640,6 @@ rsconf_systemctl_clean_unit() {
     rm -f "/etc/systemd/system/$basename$x.service"
 }
 
-rsconf_user() {
```
robnagler: Nice to delete dead code. Ideally, this would share with base_users/main.sh.jinja but later.

rsconf/package_data/opendkim/main.sh.jinja
```diff
@@ -0,0 +1,13 @@
...
+    if id -nG '{{ opendkim.socket_client_u }}' \
```
robnagler: We spell out flags when available. This code is poorly written bash:
- it uses a pipeline where bash variables would be better
- it assumes run_g isn't a part of another group name (e.g. if run_g is dkim then it would match opendkim group)
- it adds run_g to socket_client_u in a weird way instead of using base_users or another abstraction
- it assumes run_g exists, which it won't afaict.

The restart of postfix is implicit coupling. Rather, pass in the mta_service just so the build_require relationship can be controlled by the python. The comment is too detailed. Rather: `# MTA needs the new group`. 

I left this because it contains important feedback and see above.

robnagler (changes requested):

rsconf/package_data/rsconf/rsconf.sh
schellj: The problem with `flags` was that it would be an unbound variable if `gid >= 1000`. I changed it to a string to avoid having to do `${flags+"${flags[@]}"}` (which you had said annoys you :). I can do it the array way.

"system" in `system_flag` is because that's the flag to say that it's a system account (the long version being `--system`).

rsconf/package_data/opendkim/main.sh.jinja
schellj: > it assumes run_g exists, which it won't afaict.

It would be created in `rsconf_group` above, would it not?

rsconf/component/opendkim.py
schellj: > Do not allow the MTA to read the DKIM signing keys.

So you don't think this goal from the Gentoo solution is important?

rsconf/package_data/rsconf/rsconf.sh
robnagler: > `${flags+"${flags[@]}"}` (which you had said annoys you :)

yes, and it's the only way to do arrays unfortunately.

rsconf/component/opendkim.py
robnagler: There's a problem with the way this has to be done with opendkim. It doesn't offer an option to create the socket with a specific group so it has to be the process's group. This means that all that opendkim does is going to be defaulting to the socket_g on write. It should have privs to read the opendkim group. Have you tested this?

rsconf/package_data/opendkim/main.sh.jinja
robnagler: it's probably fine.

rsconf/package_data/opendkim/main.sh.jinja
robnagler: I saw that, sorry. It's just a bit odd. We don't do this other places, but I understand why it's done this way.

robnagler (changes requested):

rsconf/package_data/opendkim/main.sh.jinja
schellj: > it uses a pipeline where bash variables would be better

I'll refactor it.

> it assumes run_g isn't a part of another group name (e.g. if run_g is dkim then it would match opendkim group)

fwiw, the `-w` grep flag meant that wouldn't happen.

> it adds run_g to socket_client_u in a weird way instead of using base_users or another abstraction

What about the `usermod` call is weird? `vm_devbox` does the same thing.

rsconf/component/opendkim.py
schellj: But the process would still be running as user opendkim, so it would have read access to the keys through the owner privs. I may not be following what you mean.

I just made the effective changes in this PR manually on alpha and verified that the dkim signing worked.

rsconf/component/opendkim.py
schellj: Well, here the function needs to be called whether or not `want_unix_socket` is true. ~A better name would probably be `_setup_milter`.~ (I guess `_setup_socket` is fine, since it does that and the milter uses the socket (`inet` or `unix`).

In `build_write` we only need the additional logic if `want_unix_socket`. I could create another function, but all of the logic would be gated, with nothing in the `else`; happy to do it that way, but that doesn't read as well to me.

rsconf/component/opendkim.py
schellj: Currently the postfix component needs `milter`, though maybe it should be accessed through a method instead.

rsconf/package_data/opendkim/main.sh.jinja
robnagler: > fwiw, the -w grep flag meant that wouldn't happen.

Pedantic: `echo dkim-foo | grep -w dkim` is true. The "standard" form is `" ${id -nG ..}  " =~ " $group "`.  The `-w` slipped past me which is why we prefer long options.

These types of errors are then subtle figure out. With AI, we no longer have to take shortcuts. We can just tell it to write the code properly the first time.

> What about the usermod call is weird? vm_devbox does the same thing. 

Fair. I think it should be a function so that we aren't duplicating code and this would fix the missing check in vn_devbox, which is cleaner.

rsconf/component/opendkim.py
robnagler: Just leave it.

rsconf/component/opendkim.py
robnagler: that's fine.

rsconf/component/opendkim.py
robnagler: I was saying it is a deficiency of opendkim that it doesn't allow assigning privs to the socket. There's no good solution to this. It's better to share the socket_g than to make the socket other (world) read-write. Just go ahead with the simple solution of giving `socket _g (=opendkim) group privs to postfix and opendkim. You can restrict the read/write permissions of the /etc/opendkim to user (socket_u=opendkim) only. This is enough security.

schellj: I don't follow why `socket_g=opendkim` rather than `socket_g=opendimsocket`? The unit can be run as `opendkim:opendkimsocket` and the `postfix` user can be added to the `opendkimsocket` group that's specifically for this purpose.

robnagler (approved):
lgtm

